Issue #1440, opened 2014-07-20
M113 Ambulance not firing smoke


Markokil321:
Hello,

We used to be able to fire smoke with the m113 ambulances by pressing R whilst in the driver seat. This no longer works, it still makes the firing noise (same noise when launching flares from a heli) But no smoke appears. This was working until the 1.0.5.1/112555 update. I thought it may have something to do with an edit on our server, but it doesn't work with default epoch files either. Could this be another issue with 112555?

Effecting all: 

M113_
M113Ambul_
++ any land vehicle that is able to fire smoke

vehicles

ebaydayz:
Works great now, thanks.

PR #1514, opened 2014-08-09
Update M113.hpp

McKeighan:
Added a fixed version of the M113_DZ. It uses the TK based skin, but is set to actually work correctly.

The current vehicles in this file I believe don't work right because the turrets are being referenced incorrectly (in my opinion)

### DIFF
--- a/SQF/dayz_code/Configs/CfgVehicles/LAND/M113.hpp
+++ b/SQF/dayz_code/Configs/CfgVehicles/LAND/M113.hpp
@@ -1,9 +1,318 @@
 // Armed
-class M113_UN_EP1;
+class M113_Base: Tank {
+	brakeDistance = 9;
+  	htMin = 60;
+  	htMax = 1800;
+	afMax = 200;
+	mfMax = 100;
+	mFact = 1;
+	tBody = 250;
+	driverForceOptics = 0;
+	LODTurnedOut = 1200;
+	radarType = 0; //REMOVES RADAR FROM ALL M113 VARIENTS
+	crewVulnerable = 0;
+	viewDriverInExternal = 1;
+	tracksSpeed = 3;
+	scope = 2;
+	picture = "\ca\tracked_e\Data\ui\Picture_m113_CA.paa";
+	Icon = "\Ca\tracked_e\Data\ui\Icon_m113_CA.paa";
+	mapSize = 5;
+	displayName = "$STR_DN_M113";
+	accuracy = 0.3;
+	armor = 105;
+	damageResistance = 0.03249;
+	maxSpeed = 66;
+	hasGunner = 1;
+	hasCommander = 0;
+	viewCargoShadow = 1;
+	viewCargoShadowDiff = 0.05;
+	viewCargoShadowAmb = 0.5;
+	transportSoldier = 11;
+	irScanRangeMin = 0;
+	irScanRangeMax = 0;
+	supplyRadius = 1.8;
+	transportAmmo = 0;
+	gunnerHasFlares = 0;
+	weapons[] = {"SmokeLauncher"};
+	magazines[] = {"SmokeLauncherMag"};
+	outsideSoundFilter = 1;
+	insideSoundCoef = 0.9;
+	soundGear[] = {"",5.62341e-005,1};
+	soundGetIn[] = {"ca\SOUNDS\Vehicles\Tracked\T72\int\int-tank-diesel-door-1",0.562341,1};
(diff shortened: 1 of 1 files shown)


Windmolders:
Has this been tested?

ebaydayz:
I just tested this and noticed a few problems. On line 287 you are missing a ; It should be:
<code>displayName = "M113 DZ";</code>

On line 311 you should add this since you removed it from line 1:
<code>class M113_UN_EP1;</code>
Otherwise you get an undefined base class error. 

Once you make those two corrections it loads fine. It fixes the display names on the M113_UN_EP1_DZ and M113_TK_EP1_DZ. However, the gunner's seat is still inaccessible on both of those classes as it is currently in 1.0.5.1. The gunner's seat works fine on these classes: M113_DZ, M113_DZE, M113_PMC, M113_TK_EP1. 

I get these errors in RPT:
<pre>M113_TK_EP1_DZ: MainTurret - unknown animation source mainTurret
M113_TK_EP1_DZ: MainGun - unknown animation source mainGun
M113_TK_EP1_DZ: HatchGunner - unknown animation source hatchGunner
M113_UN_EP1_DZ: MainTurret - unknown animation source mainTurret
M113_UN_EP1_DZ: MainGun - unknown animation source mainGun
M113_UN_EP1_DZ: HatchGunner - unknown animation source hatchGunner</pre>

Spawning an M113_TK_EP1_DZE produces a flood of No entry 'bin\config.bin/CfgVehicles/M113_TK_EP1_DZE/x errors. I'm sure you can spot the problem much faster than me, so I await your response.

icomrade:
Redefining a base class is probably not a good idea. I'm going to close this because the risk outweighs the reward, and also M113_TK_EP1 and M113_UN_EP1 based off of M113_BASE, so simply referencing `class M113_Base;` would work just as well

ebaydayz:
I will test that tomorrow. The goal is to:
- Fix the gunner's seat so it is accessible on the two classes I mentioned
- Fix the smokes as mentioned in https://github.com/EpochModTeam/DayZ-Epoch/issues/1440. I think they are broken on several other classes too but I still have to test
- Fix the display names on the two classes I mentioned

If we can't fix all that it's not a big deal. The gunner's seat is only inaccessible on the M113_UN_EP1_DZ and M113_TK_EP1_DZ, so admins can still spawn the DZE variants and add ammo to them if they need to.

ebaydayz:
Upon further testing I found spawning the _DZE variants causes the No entry errors with the current 1.0.6 files even without any of @McKeighan changes. The gunner seat is also not working on either the _DZ or _DZE variants. 
<pre>Warning Message: No entry 'bin\config.bin/CfgVehicles/M113_TK_EP1_DZE/Turrets/MainTurret.minElev'.
Warning Message: '/' is not a value
Warning Message: No entry 'bin\config.bin/CfgVehicles/M113_TK_EP1_DZE/Turrets/MainTurret.maxElev'.
Warning Message: '/' is not a value
Warning Message: No entry 'bin\config.bin/CfgVehicles/M113_TK_EP1_DZE/Turrets/MainTurret.minTurn'.
Warning Message: '/' is not a value
Warning Message: No entry 'bin\config.bin/CfgVehicles/M113_TK_EP1_DZE/Turrets/MainTurret.maxTurn'.
Warning Message: '/' is not a value
Warning Message: No entry 'bin\config.bin/CfgVehicles/M113_TK_EP1_DZE/Turrets/MainTurret.maxHorizontalRotSpeed'.
Warning Message: '/' is not a value
Warning Message: No entry 'bin\config.bin/CfgVehicles/M113_TK_EP1_DZE/Turrets/MainTurret.maxVerticalRotSpeed'.
Warning Message: '/' is not a value
Warning Message: No entry 'bin\config.bin/CfgVehicles/M113_TK_EP1_DZE/Turrets/MainTurret.soundServo'.
Warning Message: Size: '/' not an array
Warning Message: No entry 'bin\config.bin/CfgVehicles/M113_TK_EP1_DZE/Turrets/MainTurret.animationSourceBody'.
Warning Message: '/' is not a value
...</pre>

Because of the tricky turrets inheritance on the M113 the easiest solution was to just copy paste the two base classes (unmodified), rename the sibling classes to have _DZ and make the changes directly to them. You can see this was the approach taken with the HMMWV as well:
https://github.com/EpochModTeam/DayZ-Epoch/blob/master/SQF/dayz_code/Configs/CfgVehicles/LAND/HMMWV.hpp

It is probably not the most efficient, but it works and all errors are gone on all classes now. All the gunner's seats are fixed too.
https://github.com/EpochModTeam/DayZ-Epoch/commit/8e8cb55bd92ab878627561c8f3773be8a81a16f2

The smokes problem is unrelated as it happens with all tanks. I will have to investigate #1440 later.

PR #1517, opened 2014-08-09
Update cfgVehicles.hpp

McKeighan:
Upated handling of class Helicopter to include animation sources and turret hitpoints.


ebaydayz:
As I mentioned in https://github.com/vbawol/DayZ-Epoch/pull/1516 this does not have the intended effect of fixing the UH1Y_DZE RPT error. If this is meant to accomplish something else please let me know. In that case you can reopen it by making the edits again on the current cfgVehicles.hpp. This has been open for so long it currently has a merge conflict.

PR #1529, opened 2014-08-15
Update Lada.hpp

McKeighan:
Added "damageResistance = 0.02099;" to the DZE2 variants.  This should let the armor perform more like a humvee then before (humvee = 0.03099).  This pushes the value of the damage resistance to about 1.5X better.

Reduced fuelcapacity on all upgrade models, as the tank is set @ 50 and not 100.   Because of the lower amount, the tank upgrades shouldn't exceed 105L (to keep it at the correct ratio). However, 150 seems like a fair compromise, 1/2 way between the calculated 105L and the previous 210L.


ebaydayz:
The damageResistance changes were already made in this commit: 
https://github.com/vbawol/DayZ-Epoch/commit/106b05f1ef62113d37202742103e8b9d54504002

Since that made this a merge conflict I went ahead and made the fuel changes in a new commit: 
https://github.com/vbawol/DayZ-Epoch/commit/0753b27341412bddd362de539df59e2afc7ad3fb

PR #1530, opened 2014-08-15
Update VWGolf.hpp

McKeighan:
Updated fuel capacity to be reasonable (like Lada change). Increased damage resistance of armor.  Corrected upgrade 1's max speed, as this upgrade actually seemed to slow the VW down from it's stock 241 speed setting.


ebaydayz:
The damageResistance change was already made by @githubmil here: https://github.com/vbawol/DayZ-Epoch/commit/1ba1cc29eb115050db464c2936dde706cf1f9ff3

Since that created a merge conflict I went ahead and made the other edits in a new commit:
https://github.com/vbawol/DayZ-Epoch/commit/05b2f15e93f94b260711062622aa79746a2309ae

PR #1538, opened 2014-08-20
Update Kamaz.hpp

McKeighan:
Updated Kamaz / KamazOpen / KamazRefuel for integration of vehicle upgrades.

Added more component requirements (then normal) to fuel tank size, armor amount, and cargo expansion.


ebaydayz:
Tested and working nicely. Thanks for your contribution.

PR #1567, opened 2014-09-13
If Crafting Item is SledgeHammer

Namindu:
If the Item being crafted is a sledgehammer then it will check to see if its in the inventory, If Yes, then it will add it to the players bag. If No, then it will add it to their main gear.

Note:
1) I have not Tested it Yet
2) This currently dose not check to see if player has a bag or not.
3) Will need to add the Text of the "cutText" to the stringtables.

### DIFF
--- a/SQF/dayz_code/actions/player_craftItem.sqf
+++ b/SQF/dayz_code/actions/player_craftItem.sqf
@@ -224,10 +224,10 @@ if (_canDo) then {
 								if ((_x == "ItemSledge") && {_x in items player}) then {
 									_bag = unitBackpack player;
 									if (!isNull _bag) then {
-										cutText [format["Object [%1] is already in your inventory. Adding to backpack!",_x], "PLAIN"];
+										cutText [format[(localize "str_epoch_player_313"),_textCreate,_countOut], "PLAIN DOWN"];
 										_bag addWeaponCargoGlobal [_x,1];
 									} else {
-										cutText [format["Object [%1] is already in your inventory. Dropping on the ground!",_x], "PLAIN"];
+										cutText [format[(localize "str_epoch_player_314"),_textCreate,_countOut], "PLAIN DOWN"];
 										_object = createVehicle ["WeaponHolder",position player,[],0,"CAN_COLLIDE"];
 										_object setVariable ["permaLoot",true];
 										_object addWeaponCargoGlobal [_x,1];
--- a/SQF/dayz_code/actions/player_craftItem.sqf
+++ b/SQF/dayz_code/actions/player_craftItem.sqf
@@ -224,10 +224,10 @@ if (_canDo) then {
 								if ((_x == "ItemSledge") && {_x in items player}) then {
 									_bag = unitBackpack player;
 									if (!isNull _bag) then {
-										cutText [format[(localize "str_epoch_player_313"),_textCreate,_countOut], "PLAIN DOWN"];
+										cutText [format[(localize "str_epoch_player_313"),_x], "PLAIN DOWN"];
 										_bag addWeaponCargoGlobal [_x,1];
 									} else {
-										cutText [format[(localize "str_epoch_player_314"),_textCreate,_countOut], "PLAIN DOWN"];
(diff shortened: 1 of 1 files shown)


SQF/dayz_code/actions/player_craftItem.sqf
```diff
@@ -221,7 +221,20 @@ if (_canDo) then {
...
 							{
-								player addWeapon _x;
+								if (_x == "ItemSledge") then {
+									_inventory = weapons player;
+									_backpack = unitBackpack player;
+									if (_x in _inventory) then {
+										cutText [format["Object [%1] already in your inventory adding to Backpack!",_x], "PLAIN"];
+										_item = _x;
+										_qty = 1;
+										_backpack addWeaponCargoGlobal [_item,_qty];
```
Windmolders: _backpack is null when no backpack is equiped. Please put this check into the code.

Windmolders:
Also quite sure this code doesn't work, because you compare a string with an object.

```sqf
_x isKindOf "ItemSledge"
````

```sqf
typeOf _x in _inventory
````

ebaydayz:
I just tested this and it only occurs when ItemSledge is already in your toolbelt. If you are holding the sledge hammer in your hands you can craft one and it will get added to your toolbelt correctly. So you should check (items player) instead of (weapons player). 

Here I made it more efficient, changed it to (items player), added the check for Null backpack and added a last resort which spawns it on the ground if the player has no backpack. 
<pre><code>if ((_x == "ItemSledge") && {_x in items player}) then {
	_bag = unitBackpack player;
	if (!isNull _bag) then {
		cutText [format["Object [%1] is already in your inventory. Adding to backpack!",_x], "PLAIN"];
		_bag addWeaponCargoGlobal [_x,1];
	} else {
		cutText [format["Object [%1] is already in your inventory. Dropping on the ground!",_x], "PLAIN"];
		_object = createVehicle ["WeaponHolder",position player,[],0,"CAN_COLLIDE"];
		_object setVariable ["permaLoot",true];
		_object addWeaponCargoGlobal [_x,1];
	};
} else {
	player addWeapon _x;
};</code></pre>

I tested and it does work. 
@Windmolders - _x is a string in this case, so it is comparing two strings. Remember addWeapon requires the weapon parameter to be a string. https://community.bistudio.com/wiki/addWeapon

Windmolders:
@ebaydayz Indeed you are correct.
Swapping the sledge as weapon or tool actually changes the item class. So it are 2 different items.

I'll accept the code when changed to ebaydayz his code.

Namindu:
Changed to Ebaydayz way.

ebaydayz:
I opened a pull request to add it to stringtable.xml

Here it is with the localized strings:

<pre><code>if ((_x == "ItemSledge") && {_x in items player}) then {
	_bag = unitBackpack player;
	if (!isNull _bag) then {
		cutText [format[(localize "str_epoch_player_313"),_x], "PLAIN"];
		_bag addWeaponCargoGlobal [_x,1];
	} else {
		cutText [format[(localize "str_epoch_player_314"),_x], "PLAIN"];
		_object = createVehicle ["WeaponHolder",position player,[],0,"CAN_COLLIDE"];
		_object setVariable ["permaLoot",true];
		_object addWeaponCargoGlobal [_x,1];
	};
} else {
	player addWeapon _x;
};</code></pre>

Namindu:
done.

ebaydayz:
<code>,_textCreate,_countOut</code>

Assuming that is a typo? It should just be _x.

Namindu:
Mybad Fixed.

ebaydayz:
The only problem I notice is the cutText is overridden by the "1 completed, missing 1 more of x" cutText, since all cutText is in the same layer by default. Rather than use a different layer which may conflict with other scripts I just made it use systemChat.

Tested in combination with https://github.com/EpochModTeam/DayZ-Epoch/pull/1667 and all is working as intended. Merged the revised version here: https://github.com/EpochModTeam/DayZ-Epoch/commit/1ade44eb3589aeb1a4576c64979348d9402a01fd

Thanks @Namindu

PR #1610, opened 2015-01-21
Fixed Black Screen on Loading

deannreid:
which made users able to walk around but not see anything but DayZ Image


RimBlock:
Did anyone find out what the underlying issue was ?.

The player_monitor.fsm usually handles the post login "endLoadingScreen;".  If people are getting stuck then for some reason the FSM is not completing.  If that is the case, other code in the FSM may also not be getting run.

PR #1622, opened 2015-02-05
darkUI fix

SplenectomY:
In player_updateGUI.sqf, _foodVal and _thirstVal are calculated differently (values closer to 1 are bad) when Dayz_Dark_UI == true. When Dayz_Dark_UI == false (old system), values closer to 0 are bad. This means that when Dayz_Dark_UI == true, health regenerates when both hunger and thirst are LOW instead of high. This fix compensates by redefining the _foodVal and _thirstVal variables for the old UI system, but a better solution might be to rewrite player_updateGUI.sqf to be more compatible.

### DIFF
--- a/SQF/dayz_code/system/player_spawn_2.sqf
+++ b/SQF/dayz_code/system/player_spawn_2.sqf
@@ -125,6 +125,10 @@ while {true} do {
 	//Hunger Effect
 	_foodVal = 		dayz_statusArray select 0;
 	_thirstVal = 	dayz_statusArray select 1;
+	if(Dayz_Dark_UI) then {
+	_foodVal = 1 - (dayz_hunger / SleepFood);
+	_thirstVal = 1 - (dayz_thirst / SleepWater);
+	};
 	if (_thirstVal <= 0) then {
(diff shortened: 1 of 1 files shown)


SplenectomY:
Also, without this fix, any mods/scripts that cause you to be completely sated (such as infistar's God mode) cause you to actually drop health as if you were starving. Further more, a brand new spawn will start at 11,980 blood instead of 12,000 because of a second of "starving" before hunger and thirst start to increase.

Windmolders:
Going to go throught the 2 files tomorrow to confirm your statements.

icomrade:
should be fixed in player_updateGUI.sqf, closing for that reason.

ebaydayz:
I think this problem still exists. With the current 1.0.6 files I spawn with 11,970 blood with dark UI enabled. With dark UI disabled I spawn with 12k blood. 

player_updateGUI.sqf hasn't been modified since July 2014 which is well before this pull request was opened. From my testing @SplenectomY's fix seems to solve it, but I'm not sure if you would rather make the change in player_updateGUI.sqf @icomrade. If not I would go ahead and merge this.

PR #1624, opened 2015-02-08
Add Toggle Slow Zombies

Mikeeeyy:
Quite a few servers have slow zombies, some for easier and less annoying PVP, some because zeds are quite glitchy when sprinting. By adding this simple option we make it easier for more servers to have slow zeds and reduce the mission file size of current servers that have it.


ebaydayz:
Whether or not you agree this should be an option, servers are going to do it anyway. Since it is so few code changes it may as well be added as a toggle variable. Thanks for your contributions Mikeeeyy.

Mikeeeyy:
Exactly, it only adds more options and modularity to the mod.

SplenectomY:
I know at least some of the servers that use slow zeds (including my own) also make them hit pretty hard. Adding a simple multiplier variable to the init.sqf (and then calling it in fn_damageHandler) wouldn't really be that hard. Thoughts?

SQF/dayz_code/system/zombie_agent.fsm
```diff
@@ -700,7 +701,7 @@ class FSM
   initState="init";
   finalStates[] =
   {
-    "End"
+    "End",
```
Windmolders: Never tested this in arma ( in php this works). BUt this doesn't create syntax error?

SQF/dayz_code/system/zombie_agent.fsm
```diff
@@ -472,7 +472,8 @@ class FSM
        "" \n
        "//Move to target" \n
        "_agent moveTo _targetPos;" \n
-       "_agent forceSpeed 10;" \n
+       "_speed = if (!isNil ""DZE_slowZombies"" && {DZE_slowZombies}) then { 2 } else { 10 };" \n
```
Windmolders: {  } those should not be needed for a true/false if i'm correct ?

SQF/dayz_code/system/zombie_agent.fsm
ebaydayz: Not sure why, but this change was made in vanilla DayZ 1.8.5 too:
https://github.com/DayZMod/DayZ/blob/Development/SQF/dayz_code/system/zombie_agent.fsm

So it definitely does not error.

Mikeeeyy: @Windmolders I just copied the export out of the FSM editor once I made the changes... I presume the editor gives you the correct format/syntax.

SQF/dayz_code/system/zombie_agent.fsm
ebaydayz: Correct, braces are only needed to switch on lazy evaluation. However, in this case lazy evaluation will be slower because (!isNil "DZE_slowZombies") is going to return true the majority of the time (since it is defined in variables.sqf). Lazy evaluation should not be used here.

Lazy evaluation should always be faster, but sometimes in SQF it is not. It is faster in <b>OR</b> cases if one of the first conditions is likely to return true often. And it is faster in <b>AND</b> cases if one of the first conditions is likely to return false often. In other cases it is slower. This is because SQF is dumb and {code} evaluates slightly slower than (condition). See: https://community.bistudio.com/wiki/Code_Optimisation#Conditions

Windmolders:
Could u pull out the lazy evaluation and i will accept the request.

ebaydayz:
On second thought it makes sense to use lazy evaluation there. I forgot it is also used to provide protection against errors. 

If "DZE_slowZombies" is nil it will exit the evaluation and prevent it from generating an undefined error by checking it as a boolean. That can happen if someone uses a custom variables.sqf (or an old variables.sqf from 1.0.5.1) where DZE_slowZombies is not defined as it should be.

Windmolders:
Correct

PR #1625, opened 2015-02-08
Persistent Name Tag Setting

Mikeeeyy:
When changing skins you'd have to select whether or not you want to display your name again, this fixed that. Unless it was intentional ofc.


ebaydayz:
Doubt it was intentional, looks like someone just forgot to set it again at the bottom of the file. Good catch.

Windmolders:
Code review OK

PR #1626, opened 2015-02-08
Fix rare occasion of gear not updating

Mikeeeyy:
getPosATL doesn't work for all scenarios, it's better to just use player and let the game work out if you're above water or not etc...


vbawol:
you are correct, merged

PR #1630, opened 2015-02-19
Should Fix Error for Undefined Base

Namindu:
ErrorMessage: File z\addons\dayz_code\Configs\CfgWeapons\Weapon\Rifle\FN_FAL_ANPVS4_DZE.hpp, line 0: /CfgWeapons.FN_FAL_ANPVS4_DZE: Undefined base class 'FN_FAL_ANPVS4'

This is taken from reg. Dayzmod. If someone would double check would help.


seb3sec:
This seems to be the FN Fal model without scope (the iron sight variant). The ANPVS4 is the NV scope which is the reason for this DZE_ version (NV toggle). But I'll try the fix when I'm home.

PR #1635, opened 2015-02-19
Update player_onPause.sqf

Mikeeeyy:


icomrade:
Saving when ESC is pressed prevents roll back of characters when force closing the game. Esc, then alt + F4 at the exact same time doesn't leave time for alt or f4 saves to work before the game terminates.

Mikeeeyy:
If you read the code the player isn't actually saved, dayz_lastCheckBit is set then it checks that diag_tickTime - dayz_lastCheckBit is more than 10, obviously 10 seconds doesn't fly by in that amount of time.

icomrade:
If the player is in extended combat then it could pose an issue, in addition to being in a trader city, dead, etc.

PR #1639, opened 2015-02-22
RedRyder added missing files

arma2WillRobinson:
It was pointed out that some important files were missing from the github. I still can't believe I missed them initially, its a good thing I don't do this for a living or i'd be homeless haha.


RimBlock:
Cheers Will.

Can you please confirm how the pbo should be built to get rid of the errors reported in the 1.0.6 build thread ?.

Thanks.

PR #1641, opened 2015-02-28
Update server_updateObject.sqf

Mikeeeyy:
Better formatting and some fixes, one major fix involving the saving of inventory. It was using format which has a limit. That could be a problem with a vehicle that can hold a lot of cargo with a ton of unique mags/weapons in it.

### DIFF
--- a/SQF/dayz_server/compile/server_updateObject.sqf
+++ b/SQF/dayz_server/compile/server_updateObject.sqf
@@ -1,16 +1,18 @@
 /*
-[_object,_type] spawn server_updateObject;
+	USAGE:
+	[_object, _type] spawn server_updateObject;
 */
+
 private ["_object","_type","_objectID","_uid","_lastUpdate","_needUpdate","_object_position","_object_inventory","_object_damage","_isNotOk","_parachuteWest","_firstTime","_object_killed","_object_repair","_isbuildable"];
 
-_object = 	_this select 0;
+_object = _this select 0;
 
-if(isNull(_object)) exitWith {
-	diag_log format["Skipping Null Object: %1", _object];
+if (isNull _object) exitWith {
+	diag_log format ["Skipping Null Object: %1", _object];
 };
 
-_type = 	_this select 1;
-_parachuteWest = ((typeOf _object == "ParachuteWest") || (typeOf _object == "ParachuteC"));
+_type = _this select 1;
+_parachuteWest = (typeOf _object) in ["ParachuteWest","ParachuteC"];
 _isbuildable = (typeOf _object) in dayz_allowedObjects;
 _isNotOk = false;
 _firstTime = false;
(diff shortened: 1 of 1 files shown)


SQF/dayz_server/compile/server_updateObject.sqf
```diff
@@ -18,173 +20,163 @@ _firstTime = false;
...
     _uid = "0";
 };
-if (!_parachuteWest && !(locked _object)) then {
-	if (_objectID == "0" && _uid == "0") then
-	{
+
+if (!_parachuteWest && !locked _object) then {
+	if (_objectID == "0" && _uid == "0") then {
 		_object_position = getPosATL _object;
     	_isNotOk = true;
```
Namindu: Could use an indent forwards

SQF/dayz_server/compile/server_updateObject.sqf
```diff
@@ -18,173 +20,163 @@ _firstTime = false;
...
 	} count _hitpoints;
-	
-	if (_objectID == "0") then {
-		_key = format["CHILD:306:%1:%2:%3:",_uid,_array,_damage];
-	} else {
-		_key = format["CHILD:306:%1:%2:%3:",_objectID,_array,_damage];
-	};
-	//diag_log ("HIVE: WRITE: "+ str(_key));
+
+	_key = "CHILD:306:" + _objectID + ":" + str _array + ":" + str _damage;
```
ebaydayz: Why did you take out the if () else here? _objectID may still equal zero right?

SQF/dayz_server/compile/server_updateObject.sqf
```diff
@@ -18,173 +20,163 @@ _firstTime = false;
...
+	{
+		_hit = [_object, _x] call object_getHit;
+		_selection = getText (configFile >> "CfgVehicles" >> typeOf _object >> "HitPoints" >> _x >> "name");
+		if (_hit > 0) then {
+			_array set [count _array, [_selection,_hit]];
+		};
+		_object setHit [_selection, _hit];
+	} count _hitpoints;
+
+	_key = "CHILD:306:" + _objectID + ":" + str _array + ":" + str _damage;
```
ebaydayz: Forgot the last ":" after str _damage. Should be ... + str _damage + ":";

SQF/dayz_server/compile/server_updateObject.sqf
```diff
@@ -18,173 +20,163 @@ _firstTime = false;
...
-		};
+	_inventory = [
+		getWeaponCargo _object,
+		getMagazineCargo _object,
+		getBackpackCargo _object
+	];
+	_previous = str (_object getVariable ["lastInventory", []]);
+	if (str _inventory != _previous) then {
+		_object setVariable ["lastInventory", _inventory];
+		_key = if (_objectID == "0") then { format ["CHILD:309:%1:", _uid] + str _inventory } else { format ["CHILD:303:%1:", _objectID] + str _inventory };
```
ebaydayz: Forgot the last ":" should be ... + str _inventory + ":"

SQF/dayz_server/compile/server_updateObject.sqf
```diff
@@ -18,173 +20,163 @@ _firstTime = false;
...
+		if (_hit > 0) then {
+			_array set [count _array, [_selection, _hit]];
+		};
 		_object setHit [_selection, _hit];
 	} count _hitpoints;
-	
-	_key = format["CHILD:306:%1:%2:%3:",_objectID,_array,_damage];
-	//diag_log ("HIVE: WRITE: "+ str(_key));
+
+	_key = "CHILD:306:" + _objectID + ":" + str _array + ":" + str _damage;
```
ebaydayz: Forgot the last ":"

ebaydayz:
Not sure if all of those variables you switched to use str () instead of format [] actually reach the 2048 character limit, but it is probably a good idea to do it anyway. Better safe than sorry. :+1:

SQF/dayz_server/compile/server_updateObject.sqf
Mikeeeyy: If you look it's using the same 306 call for the objectID and objectUID which isn't correct. Vehicles also only use the objectID, they don't use objectUID.

SQF/dayz_server/compile/server_updateObject.sqf
Mikeeeyy: My mistake, I use extDB you see.

Mikeeeyy:
Only the inventory needed it tbh, and maybe the hitpoints array. Thought I may as well do the rest while I was at it.

PR #1649, opened 2015-04-05
[FIX] Inventory wiping

Mikeeeyy:
dayz_Magazines isn't always set before syncing the character, and especially considering this is just on a loop makes it very unreliable. If you input an empty array to the server_playerSync then it will wipe your backpack and weapons aswell. Noticed this bug when players were complaining about it since I switched completely over to extDB. I've taken a look at the HiveExt.dll source and I'm pretty sure I found a check in there to see if the array is empty, which is still a hacky fix for this. I honestly wonder what the devs of DayZ were doing when they made this mod, the coding is just horrendous. I feel like making my own version of it from scratch sometimes. Rant over... sorry lol. Back on topic, using magazines player won't record the clip count in a magazine, but the only time that actually seems to be saved is when the gear dialog is closed because that's the only time you can record it using the gear commands.

### DIFF
--- a/SQF/dayz_code/system/player_spawn_2.sqf
+++ b/SQF/dayz_code/system/player_spawn_2.sqf
@@ -206,11 +206,10 @@ while {true} do {
 	//Save Checker
 	if (dayz_unsaved) then {
 		if ((time - dayz_lastSave) > DZE_SaveTime) then {
-			PVDZE_plr_Save = [player,dayz_Magazines,false,false];
+			PVDZE_plr_Save = [player,magazines player,false,false];
 			publicVariableServer "PVDZE_plr_Save";
 			dayz_unsaved = false;
 			dayz_lastSave = time;
-			dayz_Magazines = [];
 		};
 	};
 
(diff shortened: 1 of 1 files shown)


ebaydayz:
It is probably better to just check if dayz_Magazines is empty first like:
<pre><code>_dayzMags = if (count dayz_Magazines > 0) then {dayz_Magazines} else {magazines player};
PVDZE_plr_Save = [player,_dayzMags,false,false];</code></pre>

That way you don't have to worry about gear wiping if it isn't set yet and it will still save the magazine round count when the gear dialog is closed correctly.

Mikeeeyy:
player_gearSync is actually ran in unscheduled environment so that fix would be fine. Great work, I'll alter my changes.

Windmolders:
@ebaydayz Can you give a final thought on this ?

ebaydayz:
It looks fine to me. There are probably other cases that could reset it to an empty array or nil it besides a misconfigured server. It's better to be safe and double check than have the player's inventory wiped.

Windmolders:
CR OK.

PR #1656, opened 2015-04-20
Spawn trader city objects locally on each machine

ebaydayz:
Fewer networked objects = better performance. Trader city objects can be spawned locally on each machine because they do not need to take damage or spawn loot. Trader agents are still spawned on the server machine only. 

I used this script to automatically generate the compact arrays from each mission.sqf file:
http://epochmod.com/forum/index.php?/topic/34614-sqf-compact-editor-ouput/

It worked great, except the height was sometimes off for objects stacked on top of others. I needed to manually correct heights for certain objects on certain maps. I loaded up and tested these changes on all of the 25 Epoch maps to make sure all objects were in their right place except: isladuala, Panthera2, sara, FDF_Isle1_a, fapovo, Caribou and smd_sahrani_A2. Those are the only maps I could not get to work on my test server. If someone could test those maps (and reset any heights to the original mission.sqf values if necessary) that would be very helpful. 

Note admins can comment out this line in init.sqf: 
<pre><code>// Add trader city objects locally on each machine
_nil = [] execVM "\z\addons\dayz_code\compile\missions\DayZ_Epoch_11.Chernarus\mission.sqf";</code></pre>
And replace dayz_server\missions\x\mission.sqf to continue using the old globally spawned trader city objects if there are any issues. They can also use their own custom trader cities the same way.


skynetdev:
Great work, thank you

raymix:
if a locally spawned object like a wall gets destroyed, everyone will still see it, but you'll be able to walk and see trough it. Guess it does not matter for safezones much (as long as god mode is on)?

ebaydayz:
Yes, I set all of the local objects to allowDamage and enableSimulation false;

PR #1657, opened 2015-04-21
Spawn trader city objects locally on each machine

ebaydayz:
Fewer networked objects = better performance. Trader city objects can be spawned locally on each machine because they do not need to take damage or spawn loot. Trader agents are still spawned on the server machine only. 

I used this script to automatically generate the compact arrays from each mission.sqf file:
http://epochmod.com/forum/index.php?/topic/34614-sqf-compact-editor-ouput/

It worked great, except the height was sometimes off for objects stacked on top of others. I needed to manually correct heights for certain objects on certain maps. I loaded up and tested these changes on all of the 25 Epoch maps to make sure all objects were in their right place except: isladuala, Panthera2, sara, FDF_Isle1_a, fapovo, Caribou and smd_sahrani_A2. Those are the only maps I could not get to work on my test server. If someone could test those maps (and reset any heights to the original mission.sqf values if necessary) that would be very helpful. 

Note admins can comment out this line in init.sqf: 
<pre><code>// Add trader city objects locally on each machine
_nil = [] execVM "\z\addons\dayz_code\compile\missions\DayZ_Epoch_11.Chernarus\mission.sqf";</code></pre>
And replace dayz_server\missions\x\mission.sqf to continue using the old globally spawned trader city objects if there are any issues. They can also use their own custom trader cities the same way.


Windmolders:
Seems you had fun doing this. I assume this is tested correctly without syntax errors? Since it's alot to code review. @ebaydayz

ebaydayz:
On second thought, some of this is not ready to be merged. 

When I created the output it was all done with Chernarus loaded. I need to redo it for the other maps with the correct worlds loaded. The heights are slightly off for those because some positions which are water on Chernarus are land on other maps. 

This pull request was very messy too. I will clean it up into logical chunks so it is easier to review.

Issue #1658, opened 2015-05-15
Exploit in Clickactions


maddiniii:
Hi, 

whilst in Vehicle, if you combine for Example G3 Mags into M24 Mags, you will never loose the G3 Mag and also, after leaving the Vehicle you can continue doing that. You can create an infinite sorce of M24 Mags, with a single G3 Mag. 

For now i forbid clickactions completely while in vehicle, with overwriting ui_selectSlot.sqf, but that's only a temporary solution.

skynetdev:
@maddiniii  For who you write this issue? Nobody will fix that, this project is dead along time ago!

seb3sec:
It's an issue with Epoch+Overwatch: Overwatch uses the DMR magazine classes as a template for G3 magazines and 3 others. But Epoch extends the DMR magazine class with an itemaction for splitting and information what to add/remove from the player inventory. So if you're splitting a G3 mag it's removing a DMR mag (which is not available) and adds the M24 ones. I solved it by checking the _item-value in "actions/player_reloadMags.sqf" (Clientfiles) and added this:

´´´
...
_consume =  getArray (_config >> "ItemActions" >> "ReloadMag" >> "use") select 0;
_create =   getArray (_config >> "ItemActions" >> "ReloadMag" >> "output") select 0;

// EDIT by seb3sec: vil-ammo-dupe-fix
switch (_item) do {
    case "vil_20Rnd_762x51_SG": {
        _consume = "vil_20Rnd_762x51_SG";
    };
    case "vil_30Rnd_762x51_SG": {
        _consume = "vil_30Rnd_762x51_SG";
    };
    case "vil_20Rnd_762x51_G3": {
        _consume = "vil_20Rnd_762x51_G3";
    };
};

diag_log format ["reloadMag: %1, %2, %3", _item, _consume, _create];
// EDIT END

_item_ammo = gearSlotAmmoCount (uiNamespace getVariable 'uiControl');
...
´´´

Windmolders:
This will not be included in default epoch code since it's not an epoch solution. If you want to run overwatch and epoch together, you will have to overwrite this in your mission file.

PR #1665, opened 2015-07-10
#1658 seb3sec fix for splitting mags.

Windmolders:
#1658 reference


Windmolders:
@ebaydayz need a CROK before i gonna merge this :p and you are my only possible victime :p

ebaydayz:
This problem only exists when Epoch is combined with Overwatch. Unless there is going to be a merging of Epoch and Overwatch assets in the future it doesn't make sense to add it. 

Admins can just put this change in their mission if they want to run Epoch + Overwatch. They can easily overwrite player_reloadMags.sqf in their compiles.sqf with:
<pre><code>player_reloadMag = compile preprocessFileLineNumbers "custom\player_reloadMags.sqf";</code></pre>

Windmolders:
You are right.

PR #1670, opened 2015-09-12
Fix blood over maximum

Mikeeeyy:
Gives more blood, false positive on anti-hack and bad practice.


vbawol:
Thanks for the pull request, you should actually use 'min' here instead of 'max'

This is an example from A3 Epoch:
```
EPOCH_playerHunger = ((EPOCH_playerHunger + _data) min 5000) max 0;
```

Mikeeeyy:
Oh yeah, I always make that error, corrected.

PR #1673, opened 2016-02-13
Update keyboard eventhandler and add rules.sqf

ebaydayz:
Updated keyboard eventhandler to @facoptere's faster code from DayZ Mod.
Included weapon switch hotkeys (1 = primary, 2 = pistol, 3 = melee).

Added optional news/rules feed on player login from DayZ Mod. Set
dayz_enableRules in init.sqf.

I tested all the keyboard hot keys, including building, with current 1.0.6
files without any issues.


icomrade:
Looks good :thumbsup: unless there's going to be an effort to merge DayZ current code into DayZ Epoch then you can remove the references of and to dayz_onBack. If you are interested I have a private repo set up for that but it's far from done and I'm not sure it's worth it

Issty:
Nice code, but on my server this throw an error with fn_halo.sqf and I can not move (WASD) when I do halo jump. Please check it!

ebaydayz:
Fixed https://github.com/ebaydayz/DayZ-Epoch/commit/77198fb3be4c02d42443c8a090edac9051d1cf09. Thanks for catching that @Issty. The journal (i , j, diary) isn't ready yet, so I just commented it out for now. I plan to update it so it is at least functional as a stats display if nothing else. Most of the files for it are already in Epoch.

@icomrade The stuff I added in variables.sqf is the minimum needed to avoid undefined errors.  Dayz_onBack is used in player_switchWeapon.sqf. I didn't do all the work necessary to make the fast weapon swapping (no animations) functional. Currently the hotkeys only work for normal weapon switching, but I included player_switchWeapon so we can add it later. 

I don't know how many updates from DayZ Mod people want to merge into Epoch. I will do things like this keyboard event handler that are relatively easy and significantly improve performance. The new weapons, attachments and buildables would take more work, but they would be nice to add too.

Issty:
Works perfekt! Thanks!

icomrade:
When I try to merge this I get a conflict with a large amount of files, previously it was only the change log. If you remove/revert your commit 6a3f543 it would be a lot easier. I'll merge this in for you if you're having trouble after you revert that commit.

ebaydayz:
My fault, I created a merge conflict by trying to combine my description-update changes. I just made a new commit to resolve it: https://github.com/EpochModTeam/DayZ-Epoch/commit/441e9fb231a1a2653218eceec17cc9d5ae417fd3

Everything is identical, it is just fixed so the description-update stuff doesn't show as changes.

PR #1674, opened 2016-02-22
Russian localization update, safe localization

Adaptivity:
I'm not really sure if this will work but it should I think.

### DIFF
--- a/SQF/dayz_code/Configs/CfgMagazines.hpp
+++ b/SQF/dayz_code/Configs/CfgMagazines.hpp
@@ -895,10 +895,10 @@ class CfgMagazines {
 		scope = 2;
 		count = 1;
 		type = 256;
-		displayName = "Safe";
+		displayName = $STR_EPOCH_SAFE;
 		model = "\z\addons\dayz_epoch\models\safe_onside.p3d";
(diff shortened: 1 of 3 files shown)


icomrade:
your modifications to SQF/dayz_code/Configs/CfgVehicles/DZE/VaultStorage.hpp are not correct.

`displayName = "$STR_EPOCH_SAFE_LOCKED";`
should be
`displayName = $STR_EPOCH_SAFE_LOCKED;`
as well as other displayName changes in that file

displayName with a localized string (string-table definitions) are not wrapped in quotations. Other than that this should work just fine, make those changes and I'll merge this.

Adaptivity:
Hmm, I originally thought that quotes are shouldn't be used here. But I found some files where quotes are used on displayName. For example here: https://github.com/EpochModTeam/DayZ-Epoch/blob/24c8930a62e87c0cd4287beb9181f040dc72f79f/SQF/dayz_code/Configs/CfgMagazines/ItemWaterbottle.hpp. Does this means that this code is incorrect? I'll make changes in few minutes.

icomrade:
Strange, I would think it would think the game would assume they were strings, ArmA is weird. Thanks.

PR #1679, opened 2016-04-24
More localization support

Adaptivity:
Moved few more hardcoded strings to stringtable, changed few russian localization strings to better. Also, can "Mixed Gas Full" be renamed to "Full Mixed Gas Can"? Current name sounds a bit weird.
Also, stringtable seems to be holding corrupted '(apostrophes). For example this line: https://github.com/EpochModTeam/DayZ-Epoch/blob/master/SQF/dayz_code/stringtable.xml#L185 Is it normal or should be fixed?

### DIFF
--- a/SQF/dayz_code/Configs/CfgMagazines/DZE/Food.hpp
+++ b/SQF/dayz_code/Configs/CfgMagazines/DZE/Food.hpp
@@ -1,9 +1,9 @@
 class FoodBioMeat: FoodEdible {
 	scope = public;
-	displayName = "Bio Meat";
+	displayName = $STR_FOOD_NAME_BIOMEAT;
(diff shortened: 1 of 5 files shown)


ebaydayz:
Careful, you missed two  '$', I fixed it for you:
https://github.com/EpochModTeam/DayZ-Epoch/commit/6522cca7ce089507b3300e25585116f9ae09bdb9

It should be fine to change the display name. It doesn't affect anything besides the text appearance in game.

That's intended, stringtable.xml supports escaping special characters like HTML entities. You don't really need to do it unless you're nesting structured text, but it doesn't hurt:
https://community.bistudio.com/wiki/Stringtable.xml#Scripts

Adaptivity:
Sorry for missing $. I will be carefully next time.

PR #1682, opened 2016-04-25
More progress on localization support

Adaptivity:
I've made some changes to fn_selfActions.sqf and I'm not sure about them since I'm not familiar with Arma coding. I think it should work but @ebaydayz please double check it and fix it if needed. I'm also planning to make a PR (move some hardcoded stuff into stringtable) to vanilla DayZ and I hope you can review it @ebaydayz.

### DIFF
--- a/SQF/dayz_code/stringtable.xml
+++ b/SQF/dayz_code/stringtable.xml
@@ -6827,7 +6827,7 @@
 		<Key ID="str_success_gutted_zombie">
 			<English>%1 has been gutted, zombie parts are now on the carcass</English>
 			<German>%1 hat entkernt worden, zombie Teile nun auf der Karkasse sind</German>
-			<Russian>%1 был урезан, зомби части теперь на туше</Russian>
+			<Russian>Вы разделали: %1, куски плоти лежат на трупе</Russian>
 			<Spanish>%1 ha sido destripado, piezas de zombies están ahora en la carcasa</Spanish>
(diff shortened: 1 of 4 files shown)


SQF/dayz_code/stringtable.xml
```diff
@@ -13403,6 +13412,24 @@
 			<French>Votre humanité est trop %1 ce commerçant refuse de vous parler.</French>
 			<Czech>Tvá lidskost je příliš %1, tento obchodník s tebou odmítá mluvit.</Czech>
 		</Key>
+		<Key ID="STR_EPOCH_ACTIONS_HUMANITY_HIGH">
+			<English>high</English>
+			<German>high</German> -->
```
ebaydayz: You forgot to remove the closing comment <code>--></code> tag on this line and line 13419. Everything else is fine. Fix that and I'll merge it. 

BTW in the future you may want to put simple strings like "Book", "Lightbulb", "high", "low", etc. into Google Translate for the other languages. It is usually correct for single words.

Adaptivity: I personally don't like Google Translate. For example, all strings you added for russian localization were incorrect. I think real people who know exact language should translate them. But if you want I can do this for the next update. I'll fix this string in a second.

ebaydayz: I agree, for sentences and longer phrases it is usually very bad, but for single words it's usually okay.

It's fine to leave them all as English too, because at least then the formatting is done and it will be easy for a native speaker to go back and correct the stringtable later.

PR #1683, opened 2016-04-25
Update WireFence.hpp to match 1.8.7

Adaptivity:
Not sure if this will work. But shouldn't this should match vanilla?


icomrade:
Fort_RazorWire is used because it has a ghost preview and is supported with snap building. Whereas wireCat1 has neither. Ultimately the models are the same, but Fort_RazorWire is an Epoch buildable with additional features that makes wireCat1 incompatible. Closing for that reason

Adaptivity:
Yea, of course. Thanks for clarification.

PR #1686, opened 2016-04-30
Pullrequest/plot management

Bruce-LXXVI:
Added plotManagement by @DevZupa. Can be enabled/disabled by setting DZE_plotManagement.

### DIFF
--- a/SQF/dayz_code/Configs/RscDisplay/plotManagement.hpp
+++ b/SQF/dayz_code/Configs/RscDisplay/plotManagement.hpp
@@ -1,0 +1,187 @@
+
+class RscTextT
+{
+	access = 0;
+	type = 0;
+	idc = -1;
+	colorBackground[] = {0,0,0,0};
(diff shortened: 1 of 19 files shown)


icomrade:
I think this can be merged, otherwise if anyone feels differently just leave a comment. We should probably change the code to all use the same friendlies variable, and we still need to save the variable to the DB.

Bruce-LXXVI:
The "friendlies variable" discussion becomes a big issue if you think about it in a more complete manner:
* There are many "friend" lists (Tag, Plot, Door, Vehicle, Group, ...)
* How to store? (Plot is in inventory, door also I think, ...)
* What to store? (characterID vs. playerUID)

This should be discussed outside of this pull request as it touches many components.

icomrade:
True, never mind my comment about merging the variables. I think playerUID is the best way since it offers more persistence, having to re-add friends to groups or replace the plot whenever someone dies is extremely annoying and I don't really see a reason it should be like that anyway. Asking on the forums would probably be better, you're right.

PR #1689, opened 2016-05-02
more localization work

Adaptivity:
I won't be able to fix any typos today. So, review it and I'll fix them tomorrow.


Adaptivity:
Thanks for fixing it :+1: Going to localize even more stuff.

PR #1694, opened 2016-06-12
Pullrequest/door management

Bruce-LXXVI:
Hi. I added doorManagement by @DevZupa. It can be managed by variables in SQF/dayz_code/configVariables.sqf.

Also added/changed:
* Translation for Doors
* Fixed bug in FNC_check_owner_friends (there is still something wrong, since plot can be managed by everyone. But this is in fn_selfActions.sqf and does not directly affect doorManagement).

### DIFF
--- a/SQF/dayz_code/compile/A_Plot_for_Life/fn_check_owner_friends.sqf
+++ b/SQF/dayz_code/compile/A_Plot_for_Life/fn_check_owner_friends.sqf
@@ -42,14 +42,17 @@ if (DZE_permanentPlot) then {
 	
 	_playerUID = [_player] call FNC_GetPlayerUID;
 	_targetOwner = _target getVariable ["ownerPUID","0"];
-	_owner = (_playerUID == _targetOwner);
 } else {
+	_playerUID = dayz_characterID;
 	_friendlies	= _player getVariable ["friendlyTo",[]];
 	_targetOwner = _target getVariable ["CharacterID","0"];
-	_owner = (_targetOwner == dayz_characterID);
 };
 
-if (_player in _friendlies) then {
+// Check if player is owner of target object
+_owner = (_playerUID == _targetOwner);
+
+// Check if player is friends with owner of target object
+if (_playerUID in _friendlies) then {
 	_friendly = true;
 };
 
--- a/SQF/dayz_code/compile/A_Plot_for_Life/fn_check_owner_friends.sqf
+++ b/SQF/dayz_code/compile/A_Plot_for_Life/fn_check_owner_friends.sqf
@@ -49,7 +49,7 @@ if (DZE_permanentPlot) then {
 	_owner = (_targetOwner == dayz_characterID);
 };
 
-if (_targetOwner in _friendlies) then {
+if (_player in _friendlies) then {
 	_friendly = true;
 };
 
--- a/SQF/dayz_code/compile/A_Plot_for_Life/fn_check_owner_friends.sqf
+++ b/SQF/dayz_code/compile/A_Plot_for_Life/fn_check_owner_friends.sqf
@@ -42,18 +42,23 @@ if (DZE_permanentPlot) then {
 	
 	_playerUID = [_player] call FNC_GetPlayerUID;
 	_targetOwner = _target getVariable ["ownerPUID","0"];
+
+	// Check if player is a plot friend
+	if (_playerUID in _friendlies) then {
+		_friendly = true;
+	};
 } else {
 	_playerUID = dayz_characterID;
 	_friendlies	= _player getVariable ["friendlyTo",[]];
 	_targetOwner = _target getVariable ["CharacterID","0"];
+
+	// Check if target object owner is a friend of player
+	if (_targetOwner in _friendlies) then {
+		_friendly = true;
+	};
 };
 
 // Check if player is owner of target object
 _owner = (_playerUID == _targetOwner);
-
-// Check if player is friends with owner of target object
-if (_playerUID in _friendlies) then {
-	_friendly = true;
-};
 
 [_owner, _friendly]
--- a/SQF/dayz_code/compile/doorManagement/fn_check_access.sqf
+++ b/SQF/dayz_code/compile/doorManagement/fn_check_access.sqf
@@ -117,7 +117,7 @@ if(_isNearPlot) then {
 if(_targetType == "DOOR") then {
 	// determine door friends
 	// and check if player is one of them
-	_isTargetFriend = (_playerUID == _targetOwnerUID);  // Door owner is always a door friend
+	_isTargetFriend = _isOwner;  // Door owner is always a door friend
 	_targetFriends = _target getVariable ["doorfriends",[]];
 	{
 		if( (_x select 0) == _playerUID ) then { _isTargetFriend = true; };
--- a/SQF/dayz_code/compile/doorManagement/fn_check_access.sqf
+++ b/SQF/dayz_code/compile/doorManagement/fn_check_access.sqf
@@ -109,9 +109,7 @@ if(_isNearPlot) then {
 	
 	// determine plot management admins
 	// and check if player is one of them
-	{
-		if( _x == _playerUID ) then { _isPlotAdmin = true; };
-	} count DZE_PlotManagementAdmins;
+	if (_playerUID in DZE_PlotManagementAdmins) then { _isPlotAdmin = true; };
 };
 
 
--- a/SQF/dayz_code/compile/doorManagement/fn_check_access.sqf
+++ b/SQF/dayz_code/compile/doorManagement/fn_check_access.sqf
@@ -1,0 +1,146 @@
+/* 
+	Check object's ownership and friends (plot and door)
+	Original concept by RimBlock (github.com/RimBlock)
+
+	Extended by Bruce-LXXVI to support different target objects
+	and allow fine grained access policies
+	
+	Parameters: 
+	_this select 0: object - player calling this function
+	_this select 1: object - target to check ownership and friendlies of
+	
+	Returns:
+	_return select 0: bool - player is owner of target object
+	_return select 1: bool - player is friends with owner of target object
+	_return select 2: bool - player is plot owner
+	_return select 3: bool - player is plot friend
+	_return select 4: bool - player is plot management admin
+	_return select 5: bool - player is target friend
+	_return select 6: bool - player is target management admin
+	_return select 7: string - detected target type
+	
+
+	2,3,4: always false if no plot nearby
+	5,6: always false if _target is not a supported target
+*/
+
+private [
+	 "_player"						// Player, who wants to access the _target
+	,"_target"						// The target object the _player wants to access
+	,"_isOwner"						// return value
+	,"_isFriendly"					// return value
+	,"_isPlotOwner"					// return value
+	,"_isPlotFriend"				// return value
+	,"_isPlotAdmin"					// return value
+	,"_isTargetFriend"				// return value
+	,"_isTargetAdmin"				// return value
+	,"_targetType"					// return value
+	,"_targetOwnerUID"				// UID of the owner of _target
+	,"_playerUID"					// UID of the _player
+	,"_plotcheck"					// takes return value of FNC_find_plots
+	,"_isNearPlot"					// player is in plot's reach
+	,"_nearestPlot"					// plot object
+	,"_plotOwnerUID"				// plot owner's UID
+	,"_plotFriends"					// list of plot friends [["UID", "Name"], ..]
+	,"_targetFriends"				// list of target friends [["UID", "Name"], ..]
+];
+
+
+// assign parameters
+_player = _this select 0;
+_target = _this select 1;
+
+// Initialize return values
+_isOwner = false;
+_isFriendly = false;
+_isPlotOwner = false;
+_isPlotFriend = false;
+_isPlotAdmin = false;
+_isTargetFriend = false;
+_isTargetAdmin = false;
+_targetType = "";
+
+
+// determine target type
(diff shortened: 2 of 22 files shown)


oiad:
This looks great, my only thought is can we add a variable for the distance check for adding friends 
In file: SQF/dayz_code/actions/doorManagement/doorNearbyHumans.sqf 
`+_closePeople =  player nearEntities ["CAManBase", 20];`

I prefer to have it a server wide search, some admins prefer it to be the 20 meters or similar. This would also require a different syntax in doormanagment.hpp (text = $STR_EPOCH_DOORMANAGEMENT_NEARBY;) etc.

I guess this would also suit plotmanagment, too. I have both these mods as server wide listing instead of 20m.

Just my thoughts.

Bruce-LXXVI:
I'll push the distance feature to this PR later.

SQF/dayz_code/compile/A_Plot_for_Life/fn_check_owner_friends.sqf
```diff
@@ -49,7 +49,7 @@ if (DZE_permanentPlot) then {
 	_owner = (_targetOwner == dayz_characterID);
 };
 
-if (_targetOwner in _friendlies) then {
+if (_player in _friendlies) then {
```
ebaydayz: _player is an object and _friendlies is a list of strings, so this will always return false. To call this function with door management you need to merge in a separate check of the "doorfriends" variable similar to how the check of the "plotfriends" works above. If this is beyond your ability don't worry about it. Just revert this change and I'll add it in later.

I am aware Plot Management has no access control atm. Here in fn_selfActions:
https://github.com/EpochModTeam/DayZ-Epoch/blob/master/SQF/dayz_code/compile/fn_selfActions.sqf#L635

No condition is checked before the s_player_plotManagement action is added. It should be inside the IF statement directly above it. It looks like that was a mistake by icomrade when he merged A Plot for Life.

SQF/dayz_code/compile/player_unlockDoor.sqf
```diff
@@ -2,39 +2,61 @@
...
-		_display = findDisplay 41144;
 		_display closeDisplay 3000;
 		localize "STR_EPOCH_ACTIONS_16" call dayz_rollingMessages;
 	} else {
 		// get object combination
 		_objectCharacterID 	= _obj getVariable ["CharacterID","0"];
-
+		
+		if(DZE_doorManagement) then {
+			_doorOwnerID = if(DZE_plotforLife) then { dayz_selectedDoor getVariable ["ownerPUID","0"]; } else { dayz_selectedDoor getVariable ["characterID","0"]; };
```
ebaydayz: DZE_plotforLife should be DZE_permanentPlot

ebaydayz:
Also these translations are duplicates:
- STR_EPOCH_DOORMANAGEMENT_ADD use ---> STR_EPOCH_ADD
- STR_EPOCH_DOORMANAGEMENT_DELETE use --> STR_EPOCH_DELETE
- STR_EPOCH_DOORMANAGEMENT_CLOSE use --> STR_UI_CLOSE
- STR_EPOCH_DOORMANAGEMENT_NEARBY use --> STR_EPOCH_HUMANS_NEARBY (needs German added)
- STR_EPOCH_DOORMANAGEMENT_ADD_ALREADY use --> STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_ALREADYONTHELIST (needs German added)
- STR_EPOCH_DOORMANAGEMENT_ADD_MAXFRIENDS use --> STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_FRIENDLIMIT (needs German added)

Use this instead of STR_EPOCH_EYESCANNER_CANCEL (taken from A2OA table):
```html
<Key ID="STR_EPOCH_CANCEL">
			<English>Cancel</English>
			<French>Annuler</French>
			<Spanish>Cancelar</Spanish>
			<Italian>Annulla</Italian>
			<Hungarian>Mégse</Hungarian>
			<Polish>Anuluj</Polish>
			<Russian>Отмена</Russian>
			<German>Abbrechen</German>
			<Czech>Zrušit</Czech>
		</Key>
```

I recommend renaming `class EyeScanner` to `class DoorOptions`. The same goes for `STR_EPOCH_EYESCANNER_TITLE`. It is misleading to title the whole dialog "EyeScanner" because it provides buttons for door management, eye scan and manual code. Some admins may disable the EyeScanner button and just use it for manual code and friends. In Zupa's code the title is "Unlock / Lock Door" which is also not entirely accurate.

As Zupa mentioned on his thread we want a separate friends list for each door. We do not want to automatically include plotfriends or other friendlies. This allows fine grained access control, so you can for example, allow someone into the main lobby of your base, but not into the internal safe room, etc. You will need to modify either fnc_check_owner_friends or initDoorManagement and fn_selfActions to do that as I mentioned above. I can do it if you don't know how. You also forgot to give access to DZE_doorManagementAdmins.

SQF/dayz_code/compile/A_Plot_for_Life/fn_check_owner_friends.sqf
Bruce-LXXVI: Sorry, this was my fault. You are absolutely correct about _player and _friendlies. But searching the _target owner's ID (=_targetOwner) in the _friendlies list of the _target always returns true, when DZE_permanentPlot is true. Instead it should search for the _playerUID in _friendlies.

Please correct me if I am totally wrong.

SQF/dayz_code/compile/player_unlockDoor.sqf
Bruce-LXXVI: Ooops. Of course. You renamed it while I merged doorManagement in. This one I must have missed.

SQF/dayz_code/compile/A_Plot_for_Life/fn_check_owner_friends.sqf
ebaydayz: You're right there is a mistake there.

This statement should be moved inside the ELSE right below the <code>_owner = </code> line:
```sqf
if (_targetOwner in _friendlies) then {_friendly = true};
```

And then below this line:
```sqf
_owner = (_playerUID == _targetOwner);
```

Add:
```sqf
if (_playerUID in _friendlies) then {_friendly = true};
```
Then it will correctly handle DZE_permanentPlot off and on.

SQF/dayz_code/compile/A_Plot_for_Life/fn_check_owner_friends.sqf
```diff
@@ -42,14 +42,17 @@ if (DZE_permanentPlot) then {
...
-	_owner = (_playerUID == _targetOwner);
 } else {
+	_playerUID = dayz_characterID;
 	_friendlies	= _player getVariable ["friendlyTo",[]];
 	_targetOwner = _target getVariable ["CharacterID","0"];
-	_owner = (_targetOwner == dayz_characterID);
 };
 
-if (_targetOwner in _friendlies) then {
+// Check if player is owner of target object
```
Bruce-LXXVI: Moved both checks out of the "if" statement. Seems more readable to me.

ebaydayz: For DZE_permanentPlot = false; it should be checking if _targetOwner is in _friendlies. Your code is checking if dayz_characterID is in _friendlies in that case.

SQF/dayz_code/compile/A_Plot_for_Life/fn_check_owner_friends.sqf
```diff
@@ -42,18 +42,23 @@ if (DZE_permanentPlot) then {
...
+	if (_playerUID in _friendlies) then {
+		_friendly = true;
+	};
 } else {
 	_playerUID = dayz_characterID;
 	_friendlies	= _player getVariable ["friendlyTo",[]];
 	_targetOwner = _target getVariable ["CharacterID","0"];
+
+	// Check if target object owner is a friend of player
+	if (_targetOwner in _friendlies) then {
```
Bruce-LXXVI: I messed it up because it's the other way around.

Bruce-LXXVI:
The translations are changed as suggested by ebaydayz.

I renamed EyeScanner to DoorAccess. This way it es more consistent in the way like DoorManagement (=Admin) vs. DoorAccess (=User) and it is also more like in the real world. Manufacturers of such systems call them access control. What I don't understand is your remark about admins removing the "Eye Scan" button. Why should one remove the button and still let users manage a door friends list?

Bruce-LXXVI:
Added a new function FNC_check_access to allow fine grained access control. Depending on plot availability and target type it checks if the player is owner, tagged as friend by owner of target (this one does not work yet), plot owner, plot friend, plot management admin, target friend (~door friend), target management admin (~door management admin).

The returned information is then used to allow access to unlock or manage doors. Default settings in configVariables.sqf are very restrictive, actually only allowing the owner and admins to access door management:

```
// Set access rights (unlock) for doors
DZE_doorManagementAllowAccess_owner = true;
DZE_doorManagementAllowAccess_ownerFriendlies = false;
DZE_doorManagementAllowAccess_plotOwner = false;
DZE_doorManagementAllowAccess_plotFriends = false;
DZE_doorManagementAllowAccess_plotAdmins = false;
DZE_doorManagementAllowAccess_doorFriends = true;
DZE_doorManagementAllowAccess_doorAdmins = true;
// Set management rights for doors (beware that managers can add theirselves and others as door friends)
DZE_doorManagementAllowManage_owner = true;
DZE_doorManagementAllowManage_ownerFriendlies = false;
DZE_doorManagementAllowManage_plotOwner = false;
DZE_doorManagementAllowManage_plotFriends = false;
DZE_doorManagementAllowManage_plotAdmins = false;
DZE_doorManagementAllowManage_doorFriends = false;
DZE_doorManagementAllowManage_doorAdmins = true;
```

Bruce-LXXVI:
As suggested by @oiad there is now a variable to enable/disable the 10m radius for the player list. Text in dialog is not altered atm.

SQF/dayz_code/compile/doorManagement/fn_check_access.sqf
```diff
@@ -0,0 +1,146 @@
...
+_plotcheck = [_player, false] call FNC_find_plots;
+_isNearPlot = ((_plotcheck select 1) > 0);
+_nearestPlot = _plotcheck select 2;
+if(_isNearPlot) then {
+	// determine plot owner
+	// and check if player is owner of plot
+	if (DZE_permanentPlot) then {
+		_plotOwnerUID = _nearestPlot getVariable ["ownerPUID","0"];
+	} else {
+		_plotOwnerUID = _target getVariable ["characterID","0"];
```
ebaydayz: _target should be _nearestPlot here. The CharacterID of the door may not be the same as the CharacterID of the nearest plot pole.

SQF/dayz_code/compile/doorManagement/fn_check_access.sqf
```diff
@@ -0,0 +1,146 @@
...
+	{
+		if( (_x select 0) == _playerUID ) then { _isPlotFriend = true; };
+	} count _plotFriends;
+	
+	
+	// determine plot management admins
+	// and check if player is one of them
+	{
+		if( _x == _playerUID ) then { _isPlotAdmin = true; };
+	} count DZE_PlotManagementAdmins;
```
ebaydayz: This count loop should be replaced with:
```sqf
if (_playerUID in DZE_PlotManagementAdmins) then { _isPlotAdmin = true; };
```

SQF/dayz_code/compile/doorManagement/fn_check_access.sqf
```diff
@@ -0,0 +1,146 @@
...
+};
+
+
+// Process target type DOOR
+if(_targetType == "DOOR") then {
+	// determine door friends
+	// and check if player is one of them
+	_isTargetFriend = (_playerUID == _targetOwnerUID);  // Door owner is always a door friend
```
ebaydayz: Already checked this above. Can just be:
```sqf
_isTargetFriend = _isOwner;
```

SQF/dayz_code/compile/doorManagement/fn_check_access.sqf
```diff
@@ -0,0 +1,146 @@
...
+	_targetFriends = _target getVariable ["doorfriends",[]];
+	{
+		if( (_x select 0) == _playerUID ) then { _isTargetFriend = true; };
```
ebaydayz: Since you changed _playerUID into characterID when DZE_permanentPlot = false; above you need to have a separate variable like `_realPlayerUID = [_player] call FNC_GetPlayerUID;` at the top of this file. Then change this to 
```sqf
_if( (_x select 0) == _realPlayerUID ) then {
```
Otherwise you are breaking door management when DZE_permanentPlot = false;, because door management never saves characterIDs.

SQF/dayz_code/compile/doorManagement/fn_check_access.sqf
```diff
@@ -0,0 +1,146 @@
...
+	} count _targetFriends;
+	
+	
+	// determine door management admins
+	// and check if player is one of them
+	{
+		if( _x == _playerUID ) then { _isTargetAdmin = true; };
+	} count DZE_DoorManagementAdmins;
```
ebaydayz: Count loop should be replaced with:
```sqf
if (_realPlayerUID in DZE_DoorManagementAdmins) then { _isTargetAdmin = true; };
```

ebaydayz:
Also you forgot to add German to `STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_ALREADYONTHELIST`

Bruce-LXXVI:
You are right. It's not a good idea to take a well known variable and fill it with another value. So i reverted to _playerUID and _characterID.

PR #1697, opened 2016-06-19
Advanced trading fixes and modifications

oiad:
This adds combine currency option to the Advanced trading screen since @Airwaves man noticed it didn't have a way of doing it.

Updated stringtables with correct "paid" spelling and differing text for buying/selling.

This also addds proper currency checking/reporting for the log system that has been introduced, it's untidy but it works, instead of getting "user bought x for 1000 currency" in a non coins server, it will convert it to the proper 1 brief 5 10oz gold etc system (uses a new function z_calcDefaultCurrencyNoImg)

This also adds a check to see if the buy list is empty (as per the selling list) and returns text to the user informing them.

Re-arranged buy and sell so buy is on top of sell, it's a bit easier to read then.


oiad:
This is my first github usage, I apologize if i've done anything wrong, this took me about 3 hours to get to this stage, hah!

PR #1706, opened 2016-07-06
Various changes (snappoints, dooraccess and fn_selfactions)

oiad:
Changes to snappoints.hpp fixes issues with metal floors snapping to a cinder wall. Cinder to cinder will create small gaps and then metal floors won't properly line up unless you start with a metal floor then snap cinders. By making both the same snap size, this fixes this issue.

Changes to doorAccess.hpp makes it possible to unlock doors and downgrade them again. It was setting dze_lock_door to the players UID instead of the door code as it should so the conditions would not be met in fn_selfActions to give you additional access.

Changes to fn_selfActions refine access to certain points more accurately (the way they should be IMO)

1) Adds a check to see if the building item can be upgraded anymore before presenting the menu, this stops you receiving the "upgrade item" action even when the item is fully upgraded
2) Adds access checks to the maintain individual item menu action so only item owner, plot owner and plot friends will see it, same with the above upgrade item menu action, this will also stop you receiving this when you are on someone elses plot that you may not have access to (I.e it's useless)
3) Changes modular building access to be item owner, plot owner and plot friend to remove modular items.
4) Changes plot managment action menu so item owner, plot owner and plot friend can access it. I thought there could be a situation where item owner != plot owner. This can be reverted if that is never possible.
5) Changes downgrading so only item owner, plot owner and plot friends can access it. If someone has the door code that doesn't mean they should be able to remove the combo (IMO)

### DIFF
--- a/SQF/dayz_code/Configs/CfgExtra/snappoints.hpp
+++ b/SQF/dayz_code/Configs/CfgExtra/snappoints.hpp
@@ -242,8 +242,8 @@ class SnapBuilding {
 	class Cinder_DZE: FloorsWallsStairs { //All cinder walls and doors
 		points[] = {
 		{0,0,0,"Pivot"},
-		{-2.752, 0, 1.5,"Left"},
-		{2.752, 0, 1.5,"Right"},
+		{-2.64, 0, 1.5,"Left"},
+		{2.64, 0, 1.5,"Right"},
 		{0, 0, 3.37042,"Top"}
 		};
 		radius = 10;
@@ -254,8 +254,8 @@ class SnapBuilding {
 	class CinderWallHalf_Preview_DZ: Cinder_DZE {
 		points[] = {
 		{0,0,0,"Pivot"},
-		{-2.752, 0, 1.5,"Left"},
-		{2.752, 0, 1.5,"Right"},
+		{-2.64, 0, 1.5,"Left"},
+		{2.64, 0, 1.5,"Right"},
 		{0, 0, 1.5,"Top"}
 		};
 	};
@@ -268,8 +268,8 @@ class SnapBuilding {
 	class CinderWallHalf_DZ: Cinder_DZE {
 		points[] = {
 		{0,0,0,"Pivot"},
-		{-2.752, 0, 1.5,"Left"},
-		{2.752, 0, 1.5,"Right"},
+		{-2.64, 0, 1.5,"Left"},
+		{2.64, 0, 1.5,"Right"},
 		{0, 0, 1.5,"Top"}
 		};
 	};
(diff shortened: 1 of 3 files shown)


ebaydayz:
- With the exception of camonets and tank traps (right click on toolbox), 1051 did not allow anyone to remove built items except the item owner. I'm not sure why they didn't allow the plot owner. Item owner, plot owner and plot friends should be a good default. 

- 1051 allowed anyone to upgrade or downgrade if there was no plot pole nearby. If there was a plot nearby only the plot owner and plot owner's friends could upgrade or downgrade. Again item owner, plot owner and plot friends should be fine for a default.

- 1051 had no access control on maintain at all. It makes sense to only allow the item owner, plot owner and plot friends since they may be intentionally leaving the base unmaintained so it is deleted. In that case it is problematic if any random can maintain it. 


4) You are correct. It is possible _cursorTarget and the nearestPole return object from FNC_find_plots will not be the same. That was an oversight on my part. I will fix the comment on that.

Admins can always change these defaults in fn_selfActions. We could add config variables for all possible options like @Bruce-LXXVI did for doors, but there is a point when there is too many config options and it gets ridiculous. Most admins change at least one thing in fn_selfActions anyway.

ebaydayz:
Another note: this will prevent people who find modular items with no plot pole nearby from upgrading or downgrading them. I don't see that as a problem though. Someone correct us if we are wrong.

Using DZE_requirePlot = 0; won't matter, because in that case you will still be the item owner.

oiad:
@ebaydayz: I pushed these changes to my test server and deleted the plot pole at my established base, I can upgrade/downgrade and destroy buildables that were built by the previous plot pole but I can't access any doors anymore unless I manually unlock them, even though I manually placed the door and my UID should be in the worldspace.

ebaydayz:
If you mean you do not get the "Manage Door" action then the ownerPUID variable may not be getting setVariable'd on them. You can run this to check:
```sqf
systemChat (format ["Type:%1  UID:%2",typeOf cursorTarget, cursorTarget getVariable ["ownerPUID","0"]]);
```

I'll take a look when I get around to testing.

oiad:
I'm at work but I'm 99% sure the puid is getting set, I remember seeing it in the database in the world space when I was checking things.

On 11 July 2016 10:34:16 AM NZST, ebaydayz <notifications@github.com> wrote:
>If you mean you do not get the "Manage Door" action then the ownerPUID
>variable may not be getting setVariable'd on them. You can run this to
>check:
>```sqf
>systemChat (format ["Type:%1  UID:%2",typeOf cursorTarget, cursorTarget
>getVariable ["ownerPUID","0"]]);
>```
>
>I'll take a look when I get around to testing. 
>
>
>---
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/EpochModTeam/DayZ-Epoch/pull/1706#issuecomment-231614996

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

PR #1707, opened 2016-07-16
Plot for life fixes and typo fixes

oiad:
This fixes the issues I was having with APFL, Basically when downgrading
a locked door the ownerPUID was not getting transferred across so it was
getting a 0 PUID set on it once downgraded so you would then lose access
if there was no plot pole.

Now it correctly gets set and instead of doing 2 PVDZE_Obj_Swaps I am
using an if and else statement.

player_upgrade was also not sending the ownerID into the worldspace
field, this seems to differ from the original APFL script. I also added
in code so that if APFL was enabled and uid is 0 we claim the item with
players UID.

This fixes a typo @ebaydayz introduced to spam the .RPT on startup also
for the shotgun and a typo in the credits.

### DIFF
--- a/SQF/dayz_code/Configs/CfgWeapons/Rifles/MR43.hpp
+++ b/SQF/dayz_code/Configs/CfgWeapons/Rifles/MR43.hpp
@@ -9,7 +9,7 @@ class MR43_DZ : Rifle
 	
 	magazines[] =
 	{
-		2Rnd_12Gauge_Buck;		
+		2Rnd_12Gauge_Buck,
 		2Rnd_12Gauge_Slug		
 	};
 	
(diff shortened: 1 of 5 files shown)


ebaydayz:
Thanks, the typo was just copied directly from the vanilla development commit. I submitted the correction there too: https://github.com/DayZMod/DayZ/commit/c4bae975cf7df3df0e727b055a07faeedefa229c

You are correct. This is how player_upgrade should be when Plot For Life is merged with Build Vectors. 
https://github.com/strikerforce/DayzBuildVectors/blob/master/P4L_Snap_Replacements/A_Plot_for_Life/Action/player_upgrade.sqf#L185

PR #1709, opened 2016-07-23
Advanced trading changes

oiad:
This changes a few things in Advanced trading, mainly cosmetically.

- Removes all the commented out lines that were the same in defines.sqf, let me know if these are there for a reason?
- fixes the buying list to have localization support and has a total item count that works properly.
- adds total item count to the selling list and added localization for that
- cosmetic change displaying backpack info on trader to bring it in line with the other menus


oiad:
@ebaydayz: There are some stringtables changes I would like to do with advanced trading, some of the strings aren't proper english, they already have translations, if I submit a commit with the changes should I just remove the translations since they will be wrong?

ebaydayz:
I would leave the translations if they have roughly the same meaning. It's better than only having English. The translators can go back and fix their grammar later.

SQF/dayz_code/actions/AdvancedTrading/functions/z_at_changeBuySell.sqf
```diff
@@ -19,11 +19,11 @@ if(Z_Selling)then{
 	if (Z_CategoryView) then {
 		(findDisplay Z_AT_DIALOGWINDOW displayCtrl Z_AT_FILTERBUTTON) ctrlSetText localize "STR_EPOCH_VIEW";
 		(findDisplay Z_AT_DIALOGWINDOW displayCtrl Z_AT_BUYABLELIST) ctrlSetFontHeight 0.026; // Slightly bigger in category view	
-		(findDisplay Z_AT_DIALOGWINDOW displayCtrl Z_AT_BUYABLELIST) ctrlSetPosition [(0.21* safezoneW + safezoneX),(0.43 * safezoneH + safezoneY),(0.13 * safezoneW),(0.30 * safezoneH)];	
+		(findDisplay Z_AT_DIALOGWINDOW displayCtrl Z_AT_BUYABLELIST) ctrlSetPosition [(0.21* safezoneW + safezoneX),(0.41 * safezoneH + safezoneY),(0.13 * safezoneW),(0.30 * safezoneH)];
```
ebaydayz: If you want to do this please remove both ctrlSetPosition lines. Line 22 and line 16. The only purpose of them was to move the categories list down a bit so it wasn't squished against the top. But if you think it looks better like this it doesn't bother me either way.

ebaydayz: Nvm, I've got it since you are probably AFK.

PR #1718, opened 2016-08-01
Advanced trading fixes, plot/door management sorting

oiad:
Added sorting to the players online list on plot management and door management.
Removes redundant backpack full space check since the gun/mag check takes care of this already.
Added an exit after doing the allowBuying check so it doesn't repeat that stuff can't fit in your bag


ebaydayz:
If you look in calcFreeSpace you'll see _totalSpace is not the same as _allowedWeapons and _allowedMags. TotalSpace compensates for different size weapons, but the other two do not. We need the transportMaxWeapons check since some backpacks like the Czech Vest Pouch have transportMaxWeapons = 0; The _totalSpace check would attempt to buy a primary in there, since it has 12 slots. The _allowedMags check is not necessary for backpacks though. If _toBuyTotalMags exceeds _allowedMags then _totalSpace will always exceed _totalBagSlots.

Exiting if you can't buy is leaving STR_EPOCH_TRADE_CONTAINER_FULL orphaned since it will never be used. We may as well remove the else because you are right. There is no case where STR_EPOCH_TRADE_CONTAINER_FULL is used without an additional message which makes it redundant. 

I went ahead and made the advanced trading changes here: https://github.com/EpochModTeam/DayZ-Epoch/commit/ee3eb845a36ddaca45635d1bf6a1eed2d76622e1

Your door and plot management sorting is not going to match the Humans array. The Humans array is sorted based on nearEntities or playableUnits return order, so selecting the lb index will get the wrong player.

PR #1724, opened 2016-08-03
Advanced trading tidyness

oiad:
Makes all display vehicle info text the same size, it looked wierd that it was that little bit different.
Changes the bag weapon/mag full string to be in line with the vehicle version (for sanity)
Removes orphaned localization

### DIFF
--- a/SQF/dayz_code/actions/AdvancedTrading/functions/z_at_displayVehicleInfo.sqf
+++ b/SQF/dayz_code/actions/AdvancedTrading/functions/z_at_displayVehicleInfo.sqf
@@ -117,14 +117,14 @@ _wepText = "";
 if (Z_SingleCurrency) then {
 	_formattedText = format [
 	"<img image='%1' size='3' /><br />" +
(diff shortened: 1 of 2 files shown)


ebaydayz:
I intentionally made the vehicle class name and weapons list texts smaller because some long ones cut off the dialog on normal and larger interface size settings. i.e. `HMMWV_M1151_M2_DES_Base_EP1_DZ` and any vehicle with a couple different weapons like the special landrovers. 

The alternative is to decrease to 0.7 all around, but then everything is too small to read comfortably at small and very small settings. You could also increase the width of the dialog, but it is already pretty huge and I don't feel like testing everything again at all five interface size settings. 

I'll make the stringtable changes along with the other one you suggested in https://github.com/EpochModTeam/DayZ-Epoch/commit/a95925aac55380ae3cc674505b91846fc0854c89#commitcomment-18500811

ebaydayz:
On second thought we can go with 0.7 all around. The text in the list boxes is already smaller than that, so it doesn't really matter. Done here: https://github.com/EpochModTeam/DayZ-Epoch/commit/f6e10dd71c8c8cd89f433979bddf797ac1753801

PR #1725, opened 2016-08-05
Advance trading fixes

oiad:
Add more semantics changes so vehicle buying for guns/ammo/bags worked
the same as the backpack.

Checking against local vehicles doesn't work well with Advanced Trading
since a local vehicle is not singular, this changes how Advanced trading
sees your local vehicle by using dze_myVehicle that gets set when you
get into a vehicle.

This will now correctly select the last vehicle you were in, instead of
the first one in the local list.

(You could have 2 vehicles or more at a trader that you were local with
and it would always select the first in the list)

### DIFF
--- a/SQF/dayz_code/actions/AdvancedTrading/functions/z_at_allowBuying.sqf
+++ b/SQF/dayz_code/actions/AdvancedTrading/functions/z_at_allowBuying.sqf
@@ -124,17 +124,29 @@ if (_selection == 1) then { //vehicle
 	if (_allowedWeapons >= (_toBuyWeaps + _toolsToBuy)) then {
 		_check1 = true;
 	} else {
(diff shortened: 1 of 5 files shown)


ebaydayz:
"Anymore" implies the vehicle already has some weapons in it. Some vehicles can't fit any (like bicycles, etc.) so "Any" would be more accurate.

ebaydayz:
Needed to add an extra else below the `if (_inVehicle) then {` otherwise the action would get added and when you got out of the vehicle it would stay on it.

Also the isNil check was not needed when _inVehicle=true; since DZE_myVehicle will always be defined now in that case.  

Committed here: https://github.com/EpochModTeam/DayZ-Epoch/commit/c51fd11511e92e2866cff8364da69194c26c7ca2

Issue #1730, opened 2016-08-12
Destroyed objects are only removed locally using remove.sqf


nigeyuk:
When a player has removed ore from a vein and collected it from the ground, another player can still see the rubble on his screen and also collect the ore, this has been the case several times today in testing, on both chernarus and panthera.

vbawol:
https://github.com/EpochModTeam/DayZ-Epoch/blob/master/SQF/dayz_code/actions/remove.sqf#L208-L221

One simple fix maybe to just delete the destroyed vein: 
https://github.com/EpochModTeam/DayZ-Epoch/pull/1731

ebaydayz:
The rubble is already getting deleted correctly on the client that destroyed it here:
https://github.com/EpochModTeam/DayZ-Epoch/blob/master/SQF/dayz_code/actions/remove.sqf#L183

So attempting to delete it again on the same client won't do anything. 

The problem may be that simulation is disabled on those when they are spawned. So the destruction is not being synchronized. Although it should only be disabled on the server machine, so it is weird that is happening. 

@nigeyuk try with this line commented out and see if it still happens:
https://github.com/EpochModTeam/DayZ-Epoch/blob/master/SQF/dayz_server/compile/spawn_mineveins.sqf#L14

oiad:
This is the same with supply crates FYI

vbawol:
Oh ya thats right, The way Arma works is that these wreck models are actually spawned locally per player.

This explains two things:
1. If you delete the object its only deleted locally.
2. if you can set a variable then its only set on the local 'ruin' object.

This was never seen as a major issue and is most likely the reason for the added RNG logic to reduce chances of getting a payout.

oiad:
@ebaydayz i have a script I wrote to do something similar with lootboxes, You can still dupe but you need to be super quick/synchronized with the other player. This helps tremendously

icomrade:
Just make the wreck class a weaponholder and add a couple magazine slots?

icomrade:
Currently working on this, I'm trying to accomplish this by adding an event handler that adds the ore and gems when the vein wreck class is spawed, I also added 10 magazine classes to the wrecked vein. The location to access the gear is a little small, but it should be okay. The wreck classes will also have to be added to the server cleanup if they aren't already

ebaydayz:
@icomrade If you want to hop on discord we can test this in multiplayer. It will be easier to see exactly what the problem is.

[...]

PR #1731, opened 2016-08-12
remove mine after looted for #1730

vbawol:


ebaydayz:
Correct me if I'm wrong, but it should already be deleted regardless of whether it is a mine or not here:
https://github.com/EpochModTeam/DayZ-Epoch/blob/62d19b48db4a88c9d5585548e3aec76790ca40f9/SQF/dayz_code/actions/remove.sqf#L171

Sometimes the rubble is a different object than the original, so we may need to do a nearestObject "rubbleClassName" deleteVehicle.

PR #1735, opened 2016-08-14
Plot management fixes

oiad:
This hopefully adds a few more items to the maintain list (@icomrade)
Uses epoch_returnChange to accurately and tidily return change and
handle money.
Localize maintain_area.sqf for single currency servers


SQF/dayz_code/actions/plotManagement/maintain_area.sqf
```diff
@@ -11,7 +11,7 @@ s_player_maintain_area_preview = 1;
 
 _target = nearestObject [[player] call FNC_getPos,"Plastic_Pole_EP1_DZ"];
 
-_objectClasses = DZE_maintainClasses;
+_objectClasses = DZE_maintainClasses + DZE_LockableStorage;
```
ebaydayz: Safes and lockboxes do not need to be maintained. They were not required to be in 1051 and they are not in the default maintain_area.sqf either. Also the SQL event will not set damage on them.

SQF/dayz_code/init/variables.sqf
```diff
@@ -505,7 +505,7 @@ DZE_REPLACE_WEAPONS = [["Crossbow","DMR","M14_EP1","SVD","SVD_CAMO"],["Crossbow_
 DZE_LockableStorage = ["VaultStorage","VaultStorageLocked","LockboxStorageLocked","LockboxStorage"];
 DZE_LockedStorage = ["VaultStorageLocked","LockboxStorageLocked"];
 DZE_UnLockedStorage = ["VaultStorage","LockboxStorage"];
-DZE_maintainClasses = ["ModularItems","DZE_Housebase","LightPole_DZ","BuiltItems","Generator_DZ","Plastic_Pole_EP1_DZ","Fence_corrugated_DZ","CanvasHut_DZ","ParkBench_DZ","MetalGate_DZ","StickFence_DZ","DesertCamoNet_DZ","ForestCamoNet_DZ","DesertLargeCamoNet_DZ","ForestLargeCamoNet_DZ","DeerStand_DZ","Scaffolding_DZ","FireBarrel_DZ"];
+DZE_maintainClasses = ["ModularItems","DZE_Housebase","LightPole_DZ","BuiltItems","Generator_DZ","DZ_buildables","DZ_storage_base","Plastic_Pole_EP1_DZ","Fence_corrugated_DZ","CanvasHut_DZ","ParkBench_DZ","MetalGate_DZ","StickFence_DZ","DesertCamoNet_DZ","ForestCamoNet_DZ","DesertLargeCamoNet_DZ","ForestLargeCamoNet_DZ","DeerStand_DZ","Scaffolding_DZ","FireBarrel_DZ"];
```
ebaydayz: Tents and stashes have never needed to be maintained. I don't see any reason they should. The SQL event is not going to set damage on them either.

I agree that we should probably allow DZ_buildables to be maintained using the Epoch system for consistency. We will have to remove or merge the vanilla maintain requirements so that they don't have to be maintained in two separate ways.

oiad:
@ebaydayz all fixed.

ebaydayz:
[When checking if DZE_BuildingLimit has been exceeded](https://github.com/oiad/DayZ-Epoch/blob/2a1257a275b75aa4f0b14a7159803b02ad843564/SQF/dayz_code/actions/plotManagement/plotObjects.sqf#L6-L19) I think we want to count safes, lockboxes and tents in that case. We don't want players to be able to build an insane amount of safes or tents (like 300+) in a small area, That is the same amount of strain on the server as building 300 walls there.

That is how DZE_BuildingLimit is currently checked in dze_buildChecks:
https://github.com/EpochModTeam/DayZ-Epoch/blob/master/SQF/dayz_code/compile/dze_buildChecks.sqf#L59

However, I might be missing something, so if someone thinks safes, lockboxes and tents should not be counted against DZE_BuildingLimit please let us know.

The behavior I'm suggesting is:
1. Safes, lockboxes and tents do not need to be maintained. (Same as 1051)
2. Safes, lockboxes and tents should count against DZE_BuildingLimit. (Same as 1051). Actually 1051 had a bug where all object types were counted against DZE_BuildingLimit as you can see [[here](https://github.com/EpochModTeam/DayZ-Epoch/blob/e330fe17aad4daa2520b56a3cf8d175a88aff35c/SQF/dayz_code/actions/player_build.sqf#L11)].

oiad:
I do think safes etc need maintaining. They won't decay otherwise until deleted or unless you run the auto safe zeroing mysql event.

But they definitely should contribute to the building limit, everything except vehicles should imo.

ebaydayz:
Others are welcome to weigh in on this @icomrade @AirwavesMan , but it doesn't make sense to me to maintain a safe or tent. In real life they wouldn't really decay and require maintenance like a wall or fence would. There are already two SQL events to clear empty and abandoned safes and tents here: https://github.com/EpochModTeam/DayZ-Epoch/blob/master/Server%20Files/SQL/add_recommended_mysql_events.sql#L71-L87 So, IMO we should leave it as it was in 1051. Admins can change the maintain classes and their SQL events pretty easily if they want. 


Also if anyone was wondering, nearestObjects will find all classes that inherit from a parent class.

So for example:
```sqf
_list = nearestObjects [getPos player,"DZ_storage_base",45];
```
Will include all TentStorage, DomeTentStorage, DesertTentStorage, StashSmall, StashMedium, etc. There is no need to list each child class individually. However, DZE_maintainClasses and DZE_LockableStorage are arrays, not parent classes. So they still need to be added together. But "DZ_buildables" and "DZ_storage_base" are both parent classes, which is why they are being added to the final array separately.

icomrade:
I'd leave the classes that have to be maintained the same, having to pay for locked storage in your plot radius I am indifferent to (obviously it's probably better off as we're used to already).

AirwavesMan:
I think we should leave it like it was in 1051.

SQF/dayz_code/actions/maintain_area.sqf
```diff
@@ -37,6 +37,8 @@ if (_count == 0) exitWith {
 };
 
 _requirements = [];
+_option = _this select 0;
```
ebaydayz: FYI the parameters passed to scripts from addAction are always `_this select 3` so this will error. It's not the same as Zupa's which is just passing a regular array from the dialog button click. 

No worries though, I'm going to make some other changes, so I will fix it.

SmokeyBR:
whether it should be maintained or not for me goes either way, what it does matter to me and dare i say to players is that the total counts of objects in plot manager should reflect with the build limit, unless there is another way of checking how many objects are inside your plot.

ebaydayz:
You're right. Fixed, thanks.

PR #1741, opened 2016-08-16
Add missing re-combine actions for magazines

AirwavesMan:
Adding 
Mk17
Mk17 SD
M107
KSVK
L115
VSS

### DIFF
--- a/SQF/dayz_code/Configs/CfgMagazines/Magazines/762x51.hpp
+++ b/SQF/dayz_code/Configs/CfgMagazines/Magazines/762x51.hpp
@@ -70,3 +70,27 @@ class 200Rnd_762x51_M240 : 100Rnd_762x51_M240 {
 	count = 200;
 	descriptionShort = "Caliber: 7.62x51mm NATO <br/>Rounds: 200 <br/>Used in: M240, Mk 48 Mod 0";
 };
+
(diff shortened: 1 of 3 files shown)


SQF/dayz_code/Configs/CfgMagazines/Magazines/762x51.hpp
```diff
@@ -69,4 +69,28 @@ class 200Rnd_762x51_M240 : 100Rnd_762x51_M240 {
...
+	{
+		COMBINE_MAG
+	};
+};
+
+class 20Rnd_762x51_SB_SCAR : CA_Magazine
+{
+	displayName = $STR_DZ_MAG_20Rnd_762x51_SB_SCAR_NAME;
+	
+	cartridgeName = "762x51";
```
ebaydayz: Following the convention of the other configs, this should be "762x51_SB" to distinguish it from non-silenced I think.

SQF/dayz_code/stringtable.xml
```diff
@@ -10603,6 +10603,14 @@
 			<English>20Rnd. DMR mag.</English>
 			<Russian>20п. DMR маг.</Russian>
 		</Key>
```
ebaydayz: Please move all these new strings down into the dayz_epoch package since they aren't going to be in vanilla anytime soon.

PR #1747, opened 2016-08-21
Plot / door management userlist sorting

oiad:
This correctly sorts the nearby humans list in both mods, I have this
running on my 1051 server with no issues, @ebaydayz please advise if
this is not right, I've tested it pretty well and it is indeed selecting
the right person/UID

### DIFF
--- a/SQF/dayz_code/actions/doorManagement/doorAddFriend.sqf
+++ b/SQF/dayz_code/actions/doorManagement/doorAddFriend.sqf
@@ -12,9 +12,9 @@ _friends = TheDoor getVariable ["doorfriends",[]];
 _toAdd = [_friendUID,toArray _friendName];
 _inList = false;
 
-{ if ((_x  select 0) == (_toAdd select 0)) exitWith { _inList = true; }; } count _friends;
+{ if ((_x  select 0) == (_toAdd select 0)) exitWith { _inList = true; }; false } count _friends;
 if (_inList) exitWith { systemChat localize "STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_ALREADYONTHELIST"; };
-if ((count _friends) == DZE_MaxPlotFriends) exitWith { systemChat format[localize "STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_FRIENDLIMIT", DZE_MaxPlotFriends]; };
+if ((count _friends) == DZE_doorManagementMaxFriends) exitWith { systemChat format[localize "STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_FRIENDLIMIT", DZE_doorManagementMaxFriends]; };
 
 _friends set [(count _friends), _toAdd];
 TheDoor setVariable ["doorfriends", _friends, true];
--- a/SQF/dayz_code/actions/doorManagement/doorAddFriend.sqf
+++ b/SQF/dayz_code/actions/doorManagement/doorAddFriend.sqf
@@ -1,17 +1,24 @@
-private ["_pos","_plots","_inList","_friends","_toAdd"];
+private ["_inList","_toAdd","_friends","_friendUID","_friendName","_userList"];
+disableSerialization;
 
-_pos = _this select 0;
-if (_pos < 0) exitWith {};
-_toAdd = (Humans select _pos);
+_userList = (findDisplay 711195) displayCtrl 7101;
+
+_friendUID = _this select 0;
+_friendName = _userList lbText (lbCurSel _userList);
+
+if (_friendUID == "") exitWith {};
+
 _friends = TheDoor getVariable ["doorfriends",[]];
+_toAdd = [_friendUID,toArray _friendName];
 _inList = false;
-{
-	if((_x  select 0) == (_toAdd select 0)) exitWith { _inList = true; };
-} forEach _friends;
-if(_inList) exitWith { localize "STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_ALREADYONTHELIST" call dayz_rollingMessages; };
-if(count _friends >= DZE_doorManagementMaxFriends) exitWith { format[localize "STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_FRIENDLIMIT", DZE_doorManagementMaxFriends] call dayz_rollingMessages; };
-_friends set [count _friends, _toAdd];
+
+{ if ((_x  select 0) == (_toAdd select 0)) exitWith { _inList = true; }; } count _friends;
+if (_inList) exitWith { systemChat localize "STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_ALREADYONTHELIST"; };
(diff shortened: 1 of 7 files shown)


SQF/dayz_code/actions/doorManagement/doorAddFriend.sqf
```diff
@@ -1,17 +1,24 @@
...
-{
-	if((_x  select 0) == (_toAdd select 0)) exitWith { _inList = true; };
-} forEach _friends;
-if(_inList) exitWith { localize "STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_ALREADYONTHELIST" call dayz_rollingMessages; };
-if(count _friends >= DZE_doorManagementMaxFriends) exitWith { format[localize "STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_FRIENDLIMIT", DZE_doorManagementMaxFriends] call dayz_rollingMessages; };
-_friends set [count _friends, _toAdd];
+
+{ if ((_x  select 0) == (_toAdd select 0)) exitWith { _inList = true; }; } count _friends;
+if (_inList) exitWith { systemChat localize "STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_ALREADYONTHELIST"; };
+if ((count _friends) == DZE_MaxPlotFriends) exitWith { systemChat format[localize "STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_FRIENDLIMIT", DZE_MaxPlotFriends]; };
```
ebaydayz: Not sure why you changed this. Should be DZE_doorManagementMaxFriends.

SQF/dayz_code/actions/doorManagement/doorAddFriend.sqf
```diff
@@ -1,17 +1,24 @@
...
+_toAdd = [_friendUID,toArray _friendName];
 _inList = false;
-{
-	if((_x  select 0) == (_toAdd select 0)) exitWith { _inList = true; };
-} forEach _friends;
-if(_inList) exitWith { localize "STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_ALREADYONTHELIST" call dayz_rollingMessages; };
-if(count _friends >= DZE_doorManagementMaxFriends) exitWith { format[localize "STR_EPOCH_PLOTMANAGEMENT_ADDFRIEND_FRIENDLIMIT", DZE_doorManagementMaxFriends] call dayz_rollingMessages; };
-_friends set [count _friends, _toAdd];
+
+{ if ((_x  select 0) == (_toAdd select 0)) exitWith { _inList = true; }; } count _friends;
```
ebaydayz: It is possible the exitWith will not run. In that case this will not return a bool and will trigger a "Type x expected bool" RPT error. To fix that just add a bool at the end of the count loop like so:
```sqf
{ if ((_x  select 0) == (_toAdd select 0)) exitWith { _inList = true; }; false } count _friends;
```

SQF/dayz_code/actions/doorManagement/doorNearbyHumans.sqf
```diff
@@ -1,14 +1,19 @@
-private ["_closePeople","_friendUID","_friendName"];
+private ["_closePeople","_friendUID","_friendName","_userList"];
+disableSerialization;
+
+_userList = (findDisplay 711195) displayCtrl 7101;
+
+lbClear _userList;
+
+if (!DZE_doorManagementMustBeClose) then { _closePeople = playableUnits; } else { _closePeople = player nearEntities ["CAManBase", 10]; };
```
ebaydayz: It's simpler to make it:
```sqf
_closePeople = if (DZE_doorManagementMustBeClose) then { player nearEntities ["CAManBase", 10] } else { playableUnits };
```

SQF/dayz_code/actions/plotManagement/plotNearbyHumans.sqf
```diff
@@ -1,13 +1,19 @@
-private ["_closePeople","_friendUID","_friendName"];
+private ["_closePeople","_friendUID","_friendName","_userList"];
+disableSerialization;
+
+_userList = (findDisplay 711194) displayCtrl 7001;
+
+lbClear _userList;
+
+if (!DZE_plotManagementMustBeClose) then { _closePeople = playableUnits; } else { _closePeople = player nearEntities ["CAManBase", 10]; };
```
ebaydayz: The same syntax I mentioned above would be simpler here.

oiad:
@ebaydayz all sorted.

SQF/dayz_code/actions/plotManagement/plotAddFriend.sqf
```diff
@@ -15,7 +15,7 @@ _friends = _thePlot getVariable ["plotfriends",[]];
 _toAdd = [_friendUID,toArray _friendName];
 _inList = false;
 
-{ if ((_x  select 0) == (_toAdd select 0)) exitWith { _inList = true; }; } count _friends;
+{ if ((_x  select 0) == (_toAdd select 0)) exitWith { _inList = true; }; false } count _friends;
```
icomrade: What exactly does this change do?

ebaydayz: Whenever you use the { } loop syntax of the count command the script expects the final return to be a bool. In this case it should not be necessary because the exitWith should always run, so it will return a bool.

Here is an example of where it was returning a number:
https://github.com/oiad/DayZ-Epoch/blob/bc7c1879a328f54eb2686f0ee2d5120499604537/SQF/dayz_code/actions/plotManagement/plotNearbyHumans.sqf#L6-L13

The last command that runs in the count loop was lbAdd which returns a number. That would cause the error "Type number expected bool". I will add a note on the BI wiki page about that because it has probably caused a lot of confusion for people (myself included).

oiad: @ebaydayz wanted me to change it because sometimes it can fail apparently, he will be able to.explain it better

On 22 August 2016 10:13:07 AM NZST, icomrade <notifications@github.com> wrote:
>> @@ -15,7 +15,7 @@ _friends = _thePlot getVariable ["plotfriends",[]];
>>  _toAdd = [_friendUID,toArray _friendName];
>>  _inList = false;
>>  
>> -{ if ((_x  select 0) == (_toAdd select 0)) exitWith { _inList =
>true; }; } count _friends;
>> +{ if ((_x  select 0) == (_toAdd select 0)) exitWith { _inList =
>true; }; false } count _friends;
>
>What exactly does this change do?
>
>-- 
>You are receiving this because you authored the thread.
>Reply to this email directly or view it on GitHub:
>https://github.com/EpochModTeam/DayZ-Epoch/pull/1747/files/986e359a4939375e1e4228a43ce5b67baa53c9c8..5b01cd23a5d030375d84179015b0b95cf2c032ec#r75604480

-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

icomrade:
I didn't know that, surprised it's not mentioned anywhere. Good to know, thanks

ebaydayz:
I added a note and comment on the talk page that explains it more accurately:
https://community.bistudio.com/wiki/count
https://community.bistudio.com/wiki/Talk:count

In this case it is not an issue, so I will change it back.

PR #1749, opened 2016-08-22
Adding male and female indicator for skins

AirwavesMan:
Adding male and female indicator for skins and fixed some german translations.


ebaydayz:
I don't think this is necessary. There are less than a dozen female skins and they all have female names. 

The only one that is both male and female is Civilian which already has (female) after the name for the female version.

PR #1756, opened 2016-08-28
Plot management force maintain option

oiad:
This replaces the "refresh" menu option with a "force maintain" option.
Allows you to maintain the base even if it doesn't need it to either get
all your items in sync or if you are going away or similar.

Tested with:
dze_permanent plot true/false
z_singlecurrency true/false

Also fixes an undefined variable error in system_monitor that was
introduced
https://github.com/EpochModTeam/DayZ-Epoch/commit/8f58baf3f286454a9f398517dc4f1980bd4d5c8e

### DIFF
--- a/SQF/dayz_server/system/server_monitor.sqf
+++ b/SQF/dayz_server/system/server_monitor.sqf
@@ -92,6 +92,7 @@ diag_log ("HIVE: Streamed " + str(_val) + " objects");
 	_storageMoney = _x select 9;
 
 	//set object to be in maintenance mode
+	_maintenanceMode = false;
 	_maintenanceModeVars = [];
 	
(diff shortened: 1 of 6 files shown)


SQF/dayz_server/system/server_monitor.sqf
```diff
@@ -92,6 +92,7 @@ diag_log ("HIVE: Streamed " + str(_val) + " objects");
 	_storageMoney = _x select 9;
 
 	//set object to be in maintenance mode
+	_maintenanceMode = false;
```
oiad: @icomrade @ebaydayz this fixes the undefined variable issue I was having, if there's a better fix feel free to alter this

PR #1757, opened 2016-09-04
Toggle for salvaging from locked vehicles

oiad:
This adds the ability to remove salvaging from locked vehicles which helps avoid griefing
Fixes a spelling mistake for a localization string
Makes it possible to salvage 0% damaged items again, this was disabled as per
https://github.com/EpochModTeam/DayZ-Epoch/commit/3292d84b857f307571af9233dcfffe9d59a0b5e9
I've had this running on my server for a year or so with no issues (not that it doesn't affect it, i've just never seen it)
It seems to me it's rather silly you cant salvage undamaged items off a vehicle.
Removes Bicycle checking from salvage_vehicle.sqf and moves it to fn_selfactions to tidy things up, no use having a salvage menu that hangs when you can just not show the salvage menu all together.

### DIFF
--- a/SQF/dayz_code/actions/salvage_vehicle.sqf
+++ b/SQF/dayz_code/actions/salvage_vehicle.sqf
@@ -1,9 +1,7 @@
 private ["_part","_color","_vehicle","_PlayerNear","_hitpoints","_isATV","_is6WheelType","_HasNoGlassKind",
-"_6WheelTypeArray","_NoGlassArray","_NoExtraWheelsArray","_RemovedPartsArray","_damage","_cmpt","_configVeh","_damagePercent","_string","_handle","_cancel","_type","_isBicycle"];
+"_6WheelTypeArray","_NoGlassArray","_NoExtraWheelsArray","_RemovedPartsArray","_damage","_cmpt","_configVeh","_damagePercent","_string","_handle","_cancel","_type"];
 
 _vehicle = _this select 3;
-_isBicycle = _vehicle isKindOf "Bicycle";
-if (_isBicycle) exitWith {}; // No salvage for now. Bicycle wheels should not give full size tires. Also model does not update to show removed wheels.
 {dayz_myCursorTarget removeAction _x} count s_player_repairActions;s_player_repairActions = [];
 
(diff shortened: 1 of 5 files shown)


SQF/dayz_code/actions/salvage_vehicle.sqf
```diff
@@ -57,7 +55,7 @@ if (_is6WheelType) then {
 
 		//get every damaged part no matter how tiny damage is!
 		_damagePercent = str(round(_damage * 100))+"% Damage";
-		if (_damage < 1 && {_damage > 0}) then {
+		if (_damage < 1) then {
```
ebaydayz: As vbawol mentioned [[here]](https://github.com/EpochModTeam/DayZ-Epoch/commit/3292d84b857f307571af9233dcfffe9d59a0b5e9#commitcomment-18424167) there are apparently some hitpoints that do not get damaged and allow infinite removal. I don't see any obvious issue with vehicle_getHitpoints or the way the salvage scripts get the hitpoints list, so I wouldn't be surprised if it is just a problem with certain vehicle configs. Until the problem is solved it is better to just leave this as is. It is more game breaking if players can get infinite parts from certain vehicles than if they can't salvage 0% damage ones.

SQF/dayz_code/configVariables.sqf
```diff
@@ -42,6 +42,7 @@ DZE_SafeZonePosArray = []; //Prevent players in safeZones from being killed if t
 DZE_GemOccurance = [["ItemTopaz",10], ["ItemObsidian",8], ["ItemSapphire",6], ["ItemAmethyst",4], ["ItemEmerald",3], ["ItemCitrine",2], ["ItemRuby",1]]; //Sets how rare each gem is in the order shown when mining (whole numbers only)
 DZE_GodModeBaseExclude = []; //Array of object class names excluded from the god mode bases feature
 DZE_groupManagement = false; //Enable or disable group management. Enabled by default
+DZE_salvagedLocked = false; //Enable or disable salvaging of locked vehicles, useful for stopping griefing on locked vehicles.
```
ebaydayz: I assume you meant to name this salvageLocked.

PR #1762, opened 2016-09-09
Advanced trading and safe GUI changes

oiad:
This adds displaying of item/backpack/weapon/vehicle descriptions
Moves Weapons display on vehicles to a modular type, so if the vehicle
has no weapons it wont display, same with Description

Fixes an issue with safes not sounding keyclicks, at present they are
only active if you have music turned on, since I don't play with music
on (and many others) you will never hear them and when you do, the DTMF
signals they use are really obnoxious.

Replaced the sound with the one from the lockboxes, it's still audible
but very reasonable.


SQF/dayz_code/actions/AdvancedTrading/functions/z_at_displayBackpackInfo.sqf
```diff
@@ -32,15 +32,17 @@ if (isNumber (configFile >> 'CfgVehicles' >> _class >> 'transportMaxMagazines'))
 	_transportMaxMagazines  = getNumber (configFile >> 'CfgVehicles' >> _class >> 'transportMaxMagazines');
 };
 
+_displayText = "";
+_displayText = getText(configFile >> "CfgVehicles" >> _class >> "descriptionShort");
```
ebaydayz: DescriptionShort is identical to displayName for all the _DZ backpacks, so in most cases this will be redundant.

For vehicles it will definitely cut off the dialog on normal or larger interface sizes. We already have too much text there. For weapons it will sometimes cut off when the attachment and magazine lists are long like on the M4A1_DZ, etc. The only place it might fit is for magazines, but some descriptions are very long, so I would test on normal, large and very large interface sizes to make sure it doesn't cut off. If you end up using it here are some translations from the A2OA stringtable:
```
<Spanish>Descripción</Spanish>
<Italian>Descrizione</Italian>
<Hungarian>Leírás</Hungarian>
<Polish>Opis</Polish>
<Russian>Описание</Russian>
<German>Beschreibung</German>
<Czech>Popis</Czech>
```

For future reference you don't need to initialize it as ="", since it is in the private array. GetText will always return a string. If there is no entry it will return "" not nil.

oiad:
all fixed @ebaydayz

PR #1773, opened 2016-10-06
Add M240Nest_DZ to maintain classes variable.

oiad:
This makes it so M240Nest can be maintained. (Which it should be)

### DIFF
--- a/SQF/dayz_code/compile/fn_gearMenuChecks.sqf
+++ b/SQF/dayz_code/compile/fn_gearMenuChecks.sqf
@@ -21,7 +21,7 @@ if ((vehicle player) == player) then {
 				_gearSelection = getText (configFile >> "CfgVehicles" >> _vehType >> "memoryPointSupply");
 				_supplyPositionWorld = _x modelToWorld (_x selectionPosition _gearSelection);
 				if (_x isKindOf "USEC_ch53_E" || _x isKindOf "MV22") then {_dis = 25;};
-				if (locked _x && {(_supplyPositionWorld distance player) <= _dis} && {_cTarget == _x}) then {
+				if (locked _x && {(_supplyPositionWorld distance player) <= _dis}) then {
 					_exit = true;
 				};
 			} forEach _nearestObjects;
--- a/SQF/dayz_code/compile/fn_gearMenuChecks.sqf
+++ b/SQF/dayz_code/compile/fn_gearMenuChecks.sqf
@@ -21,7 +21,7 @@ if ((vehicle player) == player) then {
 				_gearSelection = getText (configFile >> "CfgVehicles" >> _vehType >> "memoryPointSupply");
 				_supplyPositionWorld = _x modelToWorld (_x selectionPosition _gearSelection);
 				if (_x isKindOf "USEC_ch53_E" || _x isKindOf "MV22") then {_dis = 25;};
-				if (locked _x && {(_supplyPositionWorld distance player) <= _dis}) then {
+				if (locked _x && {(_supplyPositionWorld distance player) <= _dis} && {_cTarget == _x}) then {
 					_exit = true;
 				};
 			} forEach _nearestObjects;
--- a/SQF/dayz_code/init/variables.sqf
+++ b/SQF/dayz_code/init/variables.sqf
@@ -528,7 +528,7 @@ DZE_REPLACE_WEAPONS = [["Crossbow","DMR","M14_EP1","SVD","SVD_CAMO"],["Crossbow_
 DZE_LockableStorage = ["VaultStorage","VaultStorageLocked","LockboxStorageLocked","LockboxStorage"];
 DZE_LockedStorage = ["VaultStorageLocked","LockboxStorageLocked"];
 DZE_UnLockedStorage = ["VaultStorage","LockboxStorage"];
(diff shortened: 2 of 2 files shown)


ebaydayz:
The root cause of the "fling" issue is the delay between pressing the gear key and cursorTarget being retrieved in fn_gearMenuChecks. Everything from the KeyDown EH firing to the onLoad of the gear display is running unscheduled, so it should be very fast. But there is always some fraction of a second delay waiting for the code to finish running.

The slowest part of the code is probably selecting from the big keyboard_keys array in the KeyDown EH. We should retrieve the cursorTarget at the very top of the keyDown EH so it registers as quickly as possible. 
https://github.com/EpochModTeam/DayZ-Epoch/blob/master/SQF/dayz_code/compile/keyboard.sqf#L2
```sqf
dayz_cursorTarget = cursorTarget; 
```
and initialize in variables.sqf on clients only:
```sqf
dayz_cursorTarget = objNull;
```

That will definitely help a little. It may even eliminate the need for this.

oiad:
During my testing I couldn't get this commit to do the exploit but after trying this morning you are right @icomrade 

I added some debug lines and the exploit is only happening when cursortarget is null so that may be something to go by

oiad:
By changing `_cTarget == _x` to `isNull _cTarget` I can't get it to do the exploit now, would that be suitable?

ebaydayz:
Also clicking the "Gear" addAction runs scheduled, so it has a much higher delay.

To fix that we can just update the variable at the top of fn_selfActions every frame too: 
```sqf
dayz_cursorTarget = cursorTarget;
```
Then use it in fn_gearMenuChecks instead of retrieving cursorTarget again.

oiad:
I reverted my change for fn_gearMenuChecks, I've got to run to work but i'm interested to see what you guys come up with.

icomrade:
If there was a way to check the object that is having the gear opened on it without cursor target that would be great, but I have yet to find one. Removing the gear action would be an obvious solution but I don't think it's one we should take.

Right now the current distance is calculated from the supply/gear point of the vehicle  so it should be much less annoying than before, the distance should be much more accurate as well

icomrade:
I already tried doing as you suggested ebay in the keydown file, it didn't work :/ Also I should note this doesn't only happen when the cursor target is null it happens any time the cursortarget isn't a locked vehicle in range.

ebaydayz:
I'll test the current solution and look into it more later. I want to make sure this issue is solved, since being locked out of your gear when you aren't looking at a vehicle is probably one of the most annoying things in 1051.

icomrade:
1051 only uses cursortarget, I don't think we can really get any more lenient than that.

ebaydayz:
Maybe we can check the line of sight to determine if the player is looking at the vehicle and nothing is in the way. Something similar to dayz_losCheck with lineIntersects/terrainIntersect:
https://github.com/EpochModTeam/DayZ-Epoch/blob/master/SQF/dayz_code/init/compiles.sqf#L260-L280

icomrade:
You can get the gear action without facing the vehicle unfortunately, I believe it's just a radius around the supply point

ndavalos:
@ebaydayz maybe you can try BIS_fnc_dirTo to determine if you're looking at the vehicle to enable the gear menu.

icomrade:
You don't have to look at the vehicle to get the gear option (see my comment immediately above yours).

Directing the player's gear menu may be the best we can do at this point - see https://github.com/EpochModTeam/DayZ-Epoch/commit/b25b7636a79b2ff65455c7e90f1655041b565773

ndavalos:
@icomrade just a thought, but have you tried adding the bounding box values to the supply point value for a little more accuracy? Might help a little more with trying to get it accurate.

icomrade:
The supply point is a proxy in the model, so commands like boundingbox and sizeof don't actually work on it unfortunately. There does seem to be a correlation between the supply position radius and the actually size of the proxy.

ndavalos:
I was thinking adding the bounding box of the vehicle in question to the supply radius actually or something similar. Kind of like how you would determine the offset for an attachto command for towing.

icomrade:
Not sure if that would work since attachto uses memory points and I think the supply position is only a proxy, briefly looking through the OA models i didn't see any indication otherwise. I would assume the bounding box property is defined in the model and not the config, editing all of the vehicles to have appropriate configs to do this is not really something I'm interested in doing.

ndavalos:
You can actually get the size of the object using the bounding box:
https://community.bistudio.com/wiki/boundingBox

Then you could take that and add the size of the object to the radius of the supply point to get a more accurate distance for accessing gear as opposed to just saying 12 or 24 depending on the class.

```
_dis = getNumber (configFile >> "CfgVehicles" >> _vehType >> "supplyRadius");
_box = boundingBoxReal _cTarget;
_pos1 = _box select 0;
_pos2 = _box select 1;
_width = abs ((_pos2 select 0) - (_pos1 select 0));
_length = abs ((_pos2 select 1) - (_pos1 select 1));
_height = abs ((_pos2 select 2) - (_pos1 select 2));
_size = ceil((_length * _width) / 2);
_dis = _dis + _size;
```

That will probably get you a proper distance. You don't have to edit any configs to do that.
https://community.bistudio.com/wiki/boundingBox

Or you can just use sizeOf:
https://community.bistudio.com/wiki/sizeOf

```
_dis = getNumber (configFile >> "CfgVehicles" >> _vehType >> "supplyRadius");
_size = sizeof (typeof cursortarget);
_dis = _dis + ceil(_size * 1.5);
```

icomrade:
The issue is that the sizeof the object has nothing to do with the supply radius size or the distance to the object where you get the gear action. So neither bounding box or sizeof will offer an accurate solution. For objects with small supply radii using either command is overkill, for example the KA52 has a supplyradius of 1.2 m and the player can only get a gear action 4.5 m away, so using the sizeof or bnoundingbox size you will get somewhere around 16 m.

The code you see on this line accounts for trends I've noticed with inaccuracy of the supplyradius definition. https://github.com/EpochModTeam/DayZ-Epoch/blob/master/SQF/dayz_code/compile/fn_gearMenuChecks.sqf#L20 It's also important to notice that the distance is not set to 12 or 24, but rather only in certain cases where the distance is too large it's set to 12 (_dis min 12), or where the object is an MV22 or CH53.

PR #1776, opened 2016-10-15
Updating german stringtable, tent fix, loot table changes

AirwavesMan:
The updates for STR_ACTIONS_2TB and STR_ACTIONS_RFROMTB are important. That fixes a very old bug. It was no really readable in german because the words were a way to long.

### DIFF
--- a/SQF/dayz_code/Configs/CfgLoot/Groups/Zombies/Worker.hpp
+++ b/SQF/dayz_code/Configs/CfgLoot/Groups/Zombies/Worker.hpp
@@ -8,12 +8,12 @@ ZombieWorker[] =
 	{Loot_MAGAZINE,		2,		ItemDocument},
 	{Loot_MAGAZINE,		2,		ItemWire},
 	{Loot_MAGAZINE,		3,		ItemTankTrap},
-	{Loot_MAGAZINE,		3,		ItemComboLock},
+	{Loot_MAGAZINE,		2,		ItemComboLock},
 	{Loot_MAGAZINE,		2,		ItemSledgeHead}
 };
 
 ZombieWorkerViral[] =
 {
-	{Loot_GROUP,		10,		ZombieCivilian},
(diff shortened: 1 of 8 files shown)


SQF/dayz_code/Configs/CfgVehicles/DZE/TentStorage.hpp
```diff
@@ -21,6 +21,7 @@ class DesertTentStorage: DomeTentStorage_base {
 	armor = 50;
 	displayname = $STR_VEH_NAME_DESERT_TENT;
 	model = "\dayz_epoch_b\models\astan.p3d";
+	pack = "WeaponHolder_ItemDesertTent";
```
AirwavesMan: Ok it's done

SQF/dayz_code/Configs/CfgLoot/Groups/Zombies/Worker.hpp
```diff
@@ -8,12 +8,12 @@ ZombieWorker[] =
...
 	{Loot_MAGAZINE,		3,		ItemTankTrap},
-	{Loot_MAGAZINE,		3,		ItemComboLock},
+	{Loot_MAGAZINE,		2,		ItemComboLock},
 	{Loot_MAGAZINE,		2,		ItemSledgeHead}
 };
 
 ZombieWorkerViral[] =
 {
-	{Loot_GROUP,		10,		ZombieCivilian},
+	{Loot_GROUP,		10,		ZombieWorker},
```
ebaydayz: Should be `ZombieWorkerViral`. Also the Suit one above should be `ZombieSuitViral`

AirwavesMan: ZombieWorker is right. because ZombieWorkerViral is the loot table im alreday in. It is the same like the other zombie loot tables.

PR #1779, opened 2016-10-26
Advanced trading coin fixes

oiad:
SC_fnc_addCoins/SC_fnc_removeCoins was not defined at all so when you purchased/sold something with Advanced Trading with coins enabled you would not get/lose any coins because these were undefined but the purchase would still go through.

I removed them completely and replaced them with the same format we use
in other scripts (maintain etc)

Also removed dayz_server.pbo from the server files directory since it's
an old 1.0.5.1 version that is no longer useful.


ndavalos:
SC_fnc_removeCoins is actually defined in zupa's coin mod, that's why it's not in the advanced trading code.

ebaydayz:
I would leave those two functions in and add a: 
```sqf
if (Z_SingleCurrency) then {
//compile them
};
```
in compiles.sqf, because they are called in other functions if single currency is installed. Otherwise people will use the non-updated versions and get errors when they try to call server_playerSync.

oiad:
@ebaydayz I reinstated those 2 functions and added the compile for them.

I think it's a bad idea to half ass things with regards to them not being defined but included in advanced trading, not everyone uses zupas version and to be honest, I assumed advanced trading was good to go straight out of the box with no addons req'd as there was no reason it wasnt/never heard otherwise.

I've kept my changes that I made to AT to keep it in line with the other stuff we have (i.e not using these 2 functions)

oiad:
Well, the problem was advanced trading was using them with no compile in sight, I'll add that compile, then.

I think people are going to assume that straight out of the box things are going to work, I did..

ndavalos:
Single currency won't work out of the box with epoch without the rest of the code for a coin mod since it's not being included. I suppose it will work enough that you'll have coins stored in the db and character, but there's no hud for it and there's no interface for storing coins on objects or in a "bank". So if someone were to implement single currency they'll either have to fix zupa's coin mod, fix their own, or write one. I rewrote zupa's coin mod, but it's customized for my own code and not in a state where it's generic enough to include otherwise. There's a bunch of stuff you have to do to get it to work properly since stuff is missing in switch model/humanity morph, safe handling, player sync etc. The framework is there for putting stuff in the db though, which goes a long way for people to implement it.

icomrade:
Honestly it might just be easier to leave all coin related files out, as it may be more work for the community if changes are needed to the added ones. There's no point in including files just for the sake of including them (Z_SingleCurrency was implemented with DB saving examples so the framework implemented in the hive would be evident), coins won't work out of the box either way. Adding coin features to the mod was meant to make it easier to upgrade the coin mod to work with the new code and hive.

ndavalos:
Yeah I would leave the hooks in there as they were, since everything was working fine the way it was once you added in a coin mod. People should just be aware of what the name of the scripts are that are called from advanced trading so they can compile their's with the same names.

oiad:
I like so far how coins has been implemented, I just thought we wouldn't include files for the sake of including them, based on how things have been kept out of the mod in the past.

I'm all for things being included and allowances being made for core mods (WAI/DZAI/DZMS etc) because at the end of the day it does benefit the mod and the community.

I definitely understood that coins wouldn't work out of the box i.e more things were required, I guess I just didn't expect Advanced Trading to not be working. The changes I have made make it stand alone now.

I closed this pull req and submitted a new one with my changes and more @icomrade @ebaydayz

PR #1787, opened 2016-11-02
Localization changes/additions.

oiad:
Fixes a few localizations, extends range for removing camo nets from
toolbox by 10m, I have found that 5m is just not enough.


SQF/dayz_code/Configs/CfgVehicles/DZE/LockboxStorage.hpp
```diff
@@ -1,6 +1,6 @@
 class LockboxStorageLocked: Land_A_tent {
 	vehicleClass = "Survival";
-	displayName = "Locked Lockbox";
+	displayName = "STR_EPOCH_LOCKBOX_LOCKED";
```
AirwavesMan: You forget to delete the " and you missed the $

SQF/dayz_code/compile/dze_buildChecks.sqf
```diff
@@ -73,5 +73,20 @@ if (_toolCheck) then {
...
+if (((count DZE_SafeZoneNoBuildItems) > 0) && {_classname in DZE_SafeZoneNoBuildItems}) then {
+	{
+		if ((player distance (_x select 0)) <  DZE_SafeZoneNoBuildDistance) exitWith { _canBuild = false; };
+	} forEach DZE_safeZonePosArray;
+	if !(_canBuild) exitWith { dayz_actionInProgress = false; format [localize "STR_EPOCH_PLAYER_166",_text,DZE_SafeZoneNoBuildDistance] call dayz_rollingMessages; [false, _isPole]; };
+};
+
+if (((count DZE_SafeZoneNoBuildNear) > 0) && {count (nearestObjects [_pos, DZE_SafeZoneNoBuildNear, DZE_BlacklistNoBuildDistance]) > 0}) exitWith {
+	dayz_actionInProgress = false;
+	format [localize "STR_EPOCH_PLAYER_167",_text,DZE_BlacklistNoBuildDistance] call dayz_rollingMessages;
```
ebaydayz: I would add a %3 to this saying what the blacklisted building type is, so the player knows if it is a barracks, control tower, etc. Just take the typeOf from the nearestObjects select 0 return.

SQF/dayz_code/configVariables.sqf
```diff
@@ -39,6 +39,10 @@ MaxAmmoBoxes = 3; // Max number of random Supply_Crate_DZE to spawn around the m
 MaxMineVeins = 50; // Max number of random mine veins to spawn around the map
 timezoneswitch = 0; // Changes murderMenu times with this offset in hours.
 DZE_SafeZonePosArray = []; //Prevent players in safeZones from being killed if their vehicle is destroyed. Format is [[[3D POS], RADIUS],[[3D POS], RADIUS]]; Ex. DZE_SafeZonePosArray = [[[6325.6772,7807.7412,0],150],[[4063.4226,11664.19,0],150]];
+DZE_SafeZoneNoBuildItems = []; // Array of object class names not allowed to build near the safe zones listed above. i.e ["VaultStorageLocked","LockboxStorageLocked","Plastic_Pole_EP1_DZ"] etc.
+DZE_SafeZoneNoBuildNear = []; //Array of object class names that are blacklisted to build near. i.e ["Land_Mil_ControlTower","Land_SS_hangar"] etc.
+DZE_SafeZoneNoBuildDistance = 1000; // Distance from safe zones listed above to disallow building near.
+DZE_BlacklistNoBuildDistance = 1000; // Distance from blacklisted objects to disallow building near.
```
ebaydayz: 1000m is too high a default for this if it is only meant to prevent blocking off certain building types with bases. For a hangar something like 150m should work for that. For a barracks you would only need it to be about 25m. Otherwise it will be confusing to see the message when there is no building of that type in sight. It will unintentionally be blocking building around a lot of other building types in dense areas. 

Admins can always increase it if they want to prevent building a camp tower next to the barracks, etc.

icomrade: It may be easier just to whitelist objects, rather than blacklisting them as well. 500 m is probably reasonable as a default, 1 Km is too high, I agree. We may just want to leave it as the size of trader areas by default, otherwise it may cause issues on smaller maps like namalsk. Also we don't even have safezones in the mod, so as eBay notes above, the name doesn't make sense

ndavalos:
I'm pretty sure he got the idea of blacklists after unpacking my server pbos and going through the code.....renaming some variables but implementing the concepts half assed as I have them implemented in my own code....

oiad:
You know you're not the only one in the world to use player distance or nearest objects? There's quite a few public scripts that use this, infistar safe zones, zupas indestructible vehicles. I've had this on my 1051 server for a while, but cheers for the benefit of the doubt

ndavalos:
You basically renamed variables and used concepts that were in my pbos after you got a copy that could be depbo'd on my own mistake of deploying a non obfuscated pbo. It has nothing to do with nearestobjects. It's really irritating, at least comrade has the courteousy to ask to use my coin conversion sql. That's a pretty lame defense there bud.

ndavalos:
Hell you even used my default blacklist distance...

oiad:
I also used a few words that are consistent with your one like "the and this"

I understand you are upset, but if you could realise you are not the only person with an idea, that would be great.

ndavalos:
It would be great if you didn't try implementing features that were not part of this mod after seeing them elsewhere and getting your hands on the code without asking the person that developed it. You obviously didn't even do it properly to begin with. Pretty douchebag move imo. Glad there's a number of things on my todo list still to correct in my code....

oiad:
There's a lot of parts of this mod that were not here originally. That's the point about this update, changing a lot of things that have been broken/needing updating for years.

I'm sorry your feels got hurt by someone that had an idea the same as everyone else, We all do things differently and we all draw inspiration from other people/ideas.

I didn't do it properly based on feedback from other epoch developers from what they want in their mod, it worked great on my server as it is, along with most of the code I submit here, it works for me but they encourage changes for the better.

icomrade:
This is Github not Maury, please bring it elsewhere.

oiad:
Changes made and pushed.

SQF/dayz_code/compile/dze_buildChecks.sqf
```diff
@@ -79,12 +79,15 @@ if (((count DZE_SafeZoneNoBuildItems) > 0) && {_classname in DZE_SafeZoneNoBuild
 	{
 		if ((player distance (_x select 0)) <  DZE_SafeZoneNoBuildDistance) exitWith { _canBuild = false; };
 	} forEach DZE_safeZonePosArray;
-	if !(_canBuild) exitWith { dayz_actionInProgress = false; format [localize "STR_EPOCH_PLAYER_166",_text,DZE_SafeZoneNoBuildDistance] call dayz_rollingMessages; [false, _isPole]; };
 };
 
-if (((count DZE_SafeZoneNoBuildNear) > 0) && {count (nearestObjects [_pos, DZE_SafeZoneNoBuildNear, DZE_BlacklistNoBuildDistance]) > 0}) exitWith {
+if !(_canBuild) exitWith { dayz_actionInProgress = false; format [localize "STR_EPOCH_PLAYER_166",_text,DZE_SafeZoneNoBuildDistance] call dayz_rollingMessages; [false, _isPole]; };
+
+_near = (nearestObjects [_pos,DZE_NoBuildNear,DZE_NoBuildNearDistance]);
```
ebaydayz: Move the search inside this check:
```sqf
if (count DZE_NoBuildNear > 0) then {
// do nearestObjects search
};
```
A nearestObjects search with an empty array is very taxing, especially greater than 50m.

oiad: @ebaydayz no worries. I'll push that when I get home from work.

SQF/dayz_code/compile/dze_buildChecks.sqf
```diff
@@ -73,5 +73,22 @@ if (_toolCheck) then {
...
+
+if !(_canBuild) exitWith { dayz_actionInProgress = false; format [localize "STR_EPOCH_PLAYER_166",_text,DZE_SafeZoneNoBuildDistance] call dayz_rollingMessages; [false, _isPole]; };
+
+if ((count DZE_NoBuildNear) > 0) then {
+	_near = (nearestObjects [_pos,DZE_NoBuildNear,DZE_NoBuildNearDistance]);
+	if ((count _near) > 0) then { _canBuild = false; };
+};
+
+if !(_canBuild) exitWith { dayz_actionInProgress = false; format [localize "STR_EPOCH_PLAYER_167",_text,DZE_NoBuildNearDistance,typeOf (_near select 0)] call dayz_rollingMessages; [false, _isPole]; };
+
```
ebaydayz: This [whole block](https://github.com/oiad/DayZ-Epoch/blob/8c0c8732a53a2e3ab52c3469a948173f22a35332/SQF/dayz_code/compile/dze_buildChecks.sqf#L76-L91) should be moved above the (_toolCheck), otherwise it is exiting with the wrong message when the has required tools check returns false.

PR #1791, opened 2016-11-05
Fixing string is not displayed correctly

AirwavesMan:
@oiad for some reason it is case sensitive here. It works only like that, otherwise there will stand $str_tear_clothes as rightclick option.


oiad:
That's weird, there are a lot of lowercase strings being used

PR #1794, opened 2016-11-10
hotwire vehicle changes

oiad:
Adds a localization for hotwire kit working
Updates localization for hotwire kit failing
Adds medic animation/sound and zombie notify


icomrade:
if you are going to play an animation prior to checking the distance from the object you can create an issue where the animation causes the player to shift backwards and causes the script to fail despite all conditions being met initially. I would move the animation inside the distance check, or vice versa.

I would assign distance to a local variable, also I'd move ```PVDZE_veh_Lock = [_vehicle,false];``` to the inside of the distance condition by making ```if ((random 10) <= 7.5) then {``` this:  ```if ((random 10) <= 7.5 && _inDistance) then {```

oiad:
@icomrade Do we even need a distance check? I don't see the point in it

icomrade:
Yeah I was trying to find one myself, I'm sure we are fine without it, most likely it was a double check to ensure the CT is actually in range

oiad:
@icomrade I've made some changes and removed the sleep.

icomrade:
Without the sleep the player may not need to wait for the anim to complete now, I'd re-add it unless it causes an issue.

oiad:
@icomrade Having the uiSleep 8; in place gives me a generic error which I can't figure out

icomrade:
Ah okay, you had me stumped because it should have worked. Thanks

PR #1797, opened 2016-11-15
Transfusion limit (To help combat humanity farming)

oiad:
This adds support for a time based transfusion limit for remote players, exactly the same as the self blood bag system.


oiad:
@ebaydayz @icomrade this is an attempt to get some thoughts on a good way of reudcing humanity farming from blood bagging your friends to gain humanity, I believe this is something we should definitely help fix before 1.0.6 goes live, this is quite an abused mechanic.

We could also extend this to morphine and so forth.

I just copied the self blood bag code and it works great but I am hoping for your thoughts if this needs anything else.

icomrade:
2 minutes should be fine, maybe a bit long since sometimes things happen so quickly, but rarely do I give more than 1 bag in 1-2 minutes

SQF/dayz_code/medical/transfusion.sqf
```diff
@@ -190,6 +194,7 @@ while {r_doLoop} do {
 	_blood = _unit getVariable ["USEC_BloodQty", 0];
 
 	if (_blood >= r_player_bloodTotal or _bloodAmount == 0) then {
+		if (DZE_transfuseLimit select 0) then { _unit setVariable ["LastTransfusion",time,true]; };
```
ebaydayz: You should set it locally on the player instead. Then you don't need to do a global setVariable and he can't go around his circle of friends and give each of them a bloodbag.

icomrade:
There shouldn't be setvariable at all, a global non-networked variable should be fine. The problem isn't 1 player getting bloodbagged by 8 people, but rather 1 person bloodbagging the same player multiple times or bloodbagging multiple players.

ebaydayz:
Yeah, a global var for "LastTransfusion" and "LastSelfTransfusion" would be better than setVariable on either player object. The setVariable can be bypassed by changing clothes or relogging (unless we reset it in humanity morph too). The global var can only be bypassed by relogging. 

Either way these are only partial solutions until we save to database, which I don't think is worth the effort to add at this late stage.

oiad:
Couldn't the same be said about the self bloodbag system then? Since that also uses set variable (I just copied that)

icomrade:
Unless you're looking to disable the action to give a player blood x amount of time since their last transfusion then we don't need SetVar (we really shouldn't do this). Other player blood is referenced by a handful of scripts, including the server, but I don't think we really need to reference this variable on any player other than the local one (the one giving the blood).

ebaydayz:
Ok, they are global vars now:
cdb9776

PR #1801, opened 2016-11-26
dze_buildchecks plot pole distance fix

oiad:
Fixes the proper distance checks for dze_buildChecks.

Thanks to @SmokeyBR https://github.com/EpochModTeam/DayZ-Epoch/issues/1766#issuecomment-263010635

### DIFF
--- a/SQF/dayz_code/compile/dze_buildChecks.sqf
+++ b/SQF/dayz_code/compile/dze_buildChecks.sqf
@@ -15,13 +15,19 @@ if (isNumber (configFile >> "CfgVehicles" >> _classname >> "requireplot")) then 
 _isPole = (_classname == "Plastic_Pole_EP1_DZ");
 _isLandFireDZ = (_classname == "Land_Fire_DZ");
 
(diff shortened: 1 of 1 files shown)


icomrade:
Can you double check that this is correct ```_isPole = (_classname == "Plastic_Pole_EP1_DZ");``` I think that _className will be the plot pole kit magazine class, but Plastic_Pole_EP1_DZ is the vehicle class

thanks

oiad:
That is working as expected, this change definitely fixes the issues @SmokeyBR was explaining in the issues thread.

ebaydayz:
The _classname var is correct from the magazine > ItemActions > create:
https://github.com/EpochModTeam/DayZ-Epoch/blob/master/SQF/dayz_code/Configs/CfgMagazines/DZE/Misc.hpp#L378-L394

It was just always using the (DZE_PlotPole select 0) distance for the near plot check, because it was always calling fn_find_plots with false, regardless of the item being built.

PR #1820, opened 2016-12-11
Add missing guns to server traders

TheZog:
I went through the list of guns in [dayz_code/Configs/CfgLoot/Groups/Weapons.hpp](DayZ-Epoch/SQF/dayz_code/Configs/CfgLoot/Groups/Weapons.hpp) and added several guns that have DB server trader entries, but weren't found in the CfgServerTrader files.
The following guns are still missing from the CfgServerTrader files as I wasn't sure how they should be priced or which traders should buy/sell them:

AKS_GOLD
SA58_ACOG_DZ
SA58_CCO_DZ
SA58_RIS_DZ
M107_DZ


oiad:
I would add the AKS_GOLD to black market trader, the rest of the SA58's to the normal or black market trader. Price wise, I would keep them the same as what is near it in terms of fire power etc.

The M107 I would leave out, leave it up to the admin to add that one (same as AS50)

icomrade:
You are missing a semi-colon  https://github.com/EpochModTeam/DayZ-Epoch/pull/1820/commits/b3f9fb8f5faad68c7aa4fb5a9f2e1e28de29501a#diff-4f348fca69c7cc71a66d2134013e811dR146

TheZog:
I've corrected the missing semicolon and added the rest of the guns (minus the M107_DZ) in a second commit. I believe that the AKS_GOLD is only found on the Lingor map, I've added it to the bandit trader and just sort of guessed a price for it based on some of the other gun prices.

ebayShopper:
I intentionally did not add variants with attachments already on, because the rifle buy list gets way too long with all of them. For example, if we add all the AK, G36, M4 and M16 _DZ classes with every possible combination of flashlight, scope, silencer and grenade launcher that is over 80 additional weapons. Just adding the base AKs, G36, M4 and M16 keeps the list at a reasonable length. Then people can buy attachments they want separately. Every trader that sells weapons has all the attachments.

However, I guess it makes sense to add some of the most commonly bought variants along with the pistol, sniper and LMG to the buy lists.

Also now that we have the sell only option it makes sense to add all variants to the sell lists. This way players can sell without removing all attachments first. But it is important to make sure the weapon prices line up with the attachment prices, so no profit can be made by adding or removing attachments. 

With the prices you have right now a profit can be made by buying the weapon, removing the attachment and then selling both separately.

TheZog:
I'm certainly fine with not having 80+ variations and I wasn't aware of using a sell only option, but that makes sense too. I can go through the guns in my list and fix the prices for the few with attachments to be (base gun+attachment=$). I'd rather keep the game balanced like you mentioned.

For my own edification, do you have a list of all of the new attachment options with their matching guns? The main Epoch wiki seems out of date so I'm not sure where to find that info outside of digging through the code itself. 

We use [DZAI](https://github.com/SnarkIndustries/DZAI) on our server and its out-of-date code tends to use the old Dayz1.7 names when auto building a list of weapons (I've since changed its config to use static lists with the new names instead). That's when I started noticing the gun name changes and missing items at the traders; especially for guns that are in the SQL but not the cfg files.

Sorry if my pull request was a little messy. I've not really ever used GitHub or contributed before. I thought if anything, I could help get the fixes I thought might be universal into the upstream code here, rather than just posting an issue. Maybe save you a little time and effort. :)

BTW thanks for the [WAI](https://github.com/ebayShopper/WICKED-AI) fork and update! Let me know if you ever do the same to DZAI as I'd love to use it! :D

ebayShopper:
All the new _DZ classes are found here in their respective folder\file:
https://github.com/EpochModTeam/DayZ-Epoch/tree/master/SQF/dayz_code/Configs/CfgWeapons

TheZog:
I've spent some time to get more familiar with using gun attachments while in game, so I'll cancel this pull request since it's not really necessary.

PR #1823, opened 2016-12-15
Adding back the 20 stanag to 30 stanag change for M4SPR

AirwavesMan:


ebayShopper:
If you want the SPR to always spawn with 30Rnd magazines we can just overwrite the config to put the 30Rnd magazine first in the array. Or we can add an M4SPR_DZ class that does that.

AirwavesMan:
I think adding a DZE class is the best variant for that. Im not sure but is that enough for the config side?

`class M4SPR;
class M4SPR_DZE: M4SPR {
	magazines[] =
	{
		30Rnd_556x45_Stanag,
		30Rnd_556x45_StanagSD,
		20Rnd_556x45_Stanag
	};
};`

ebayShopper:
Added f1b991a

PR #1828, opened 2016-12-22
Detection of part doesn't work

ndavalos:
For tanks and other vehicles not explicitly defined in configs, _part will return with "" using gettext(), evaluating if it is nil doesn't do anything. With this change it will properly allow for repairing tracks on tanks with scrap metal.


ebayShopper:
Should be:
```sqf
if ((isNil "_part") || {_part == ""}) then {
```
Otherwise it would error if _part was undefined.

ndavalos:
_part will actually never be undefined since if gettext() doesn't find anything it returns an empty string. I fixed this locally since no one could repair tracks on tanks and it works fine.

ebayShopper:
Ok, you can just remove the isNil check then.

PR #1829, opened 2016-12-25
Move more debug lines to #ifdef

oiad:
Tidys up server RPT

Also make the combination rolling text overwrite the previous build
spam.


SQF/dayz_server/compile/server_publishFullObject.sqf
```diff
@@ -12,7 +12,9 @@ _fuel =			_this select 7;
 _allowed = [_object, "Server"] call check_publishobject;
 if (!_allowed) exitWith { deleteVehicle _object; };
```
ebayShopper: You need to add this include at the top of files that don't already have it or the macros won't work:
```sqf
#include "\z\addons\dayz_server\compile\server_toggle_debug.hpp"
```

oiad: Done

PR #1830, opened 2016-12-26
 Craft with all Dayz_Ignators

AirwavesMan:
Adding the ability to craft with all Dayz_Ignators when dayz_matchboxCount = true; For example crafting a Firebarrel


oiad:
That variable name (dayz_ignators) is horribly misspelt (not your cause)

should really be dayz_igniters

PR #1847, opened 2017-01-02
Add harder penalties for wrong combination on doors, More spelling mistakes.

oiad:
This adds a variable (DZE_doorManagementHarderPenalty) to make attempts at code breaking doors harder. This will exponentially increment the time between attempts or if you set it to false will make it 5 seconds between attempts.

Unlike the original (that was useless TBH) this will actually lock out the combo lock UI until the timer is set.

Will reset completely after 120 seconds.

@ebayShopper  Further to commit:
https://github.com/EpochModTeam/DayZ-Epoch/commit/0e938ba8f36b2631b98da1bdbfd3d6d578432110

### DIFF
--- a/SQF/dayz_code/stringtable.xml
+++ b/SQF/dayz_code/stringtable.xml
@@ -6598,7 +6598,7 @@
 			<French>Vous ne pouvez pas replier cette tente, elle ne vous appartient pas.</French>
 			<Czech>Tento stan nemůžete zabalit, jelikož není váš.</Czech>
 		</Key>
-		<Key ID="str_success_tent_destoryed">
+		<Key ID="str_success_tent_destroyed">
 			<English>You have successfully destroyed a tent.</English>
 			<German>Du hast das Zelt erfolgreich zerstört.</German>
 			<Russian>Вы успешно уничтожили палатку.</Russian>
--- a/SQF/dayz_code/stringtable.xml
+++ b/SQF/dayz_code/stringtable.xml
@@ -11280,7 +11280,7 @@
 		</Key>
 		<Key ID="STR_EPOCH_PLAYER_19">
 			<English>You entered the wrong code, wait %1 seconds.</English>
-			<German>Es wurde zu oft der falsche Code eingetippt. Du musst nun 10 Sekunden warten</German>
+			<German>Du hast einen falschen Code eingegeben, warte %1 Sekunden.</German>
 			<Russian>Неправильный код введён слишком много раз, подождите 10 секунд</Russian>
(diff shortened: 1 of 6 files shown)


Clanwarfare:
NICE!!! I could totally use this for my server! We have players that would spend all day brute forcing codes to raid bases. Going to test it now

SQF/dayz_code/stringtable.xml
```diff
@@ -11279,13 +11279,16 @@
...
+			<English>You entered the wrong code, wait %1 seconds.</English>
 			<German>Es wurde zu oft der falsche Code eingetippt. Du musst nun 10 Sekunden warten</German>
 			<Russian>Неправильный код введён слишком много раз, подождите 10 секунд</Russian>
 			<Dutch>Je hebt te vaak een foute code ingevoerd. 10 seconden geduld svp.</Dutch>
 			<French>Trop d'erreurs avec le code, patientez 10 secondes</French>
 			<Czech>Mnohokrát jsi zadal špatný kód, musíš počkat 10 vteřin</Czech>
 		</Key>
+		<Key ID="STR_EPOCH_PLAYER_19_WAIT">
+			<English>You must wait another %1 seconds before you can enter another code.</English>
+		</Key>
```
AirwavesMan: For the german string:

Du musst weitere %1 Sekunden warten, bevor du einen anderen Code eingeben kannst.

SQF/dayz_code/stringtable.xml
```diff
@@ -11279,13 +11279,16 @@
 			<Czech>Musíte mít syrové maso</Czech>
 		</Key>
 		<Key ID="STR_EPOCH_PLAYER_19">
-			<English>Wrong code entered too many times, wait 10 seconds</English>
+			<English>You entered the wrong code, wait %1 seconds.</English>
 			<German>Es wurde zu oft der falsche Code eingetippt. Du musst nun 10 Sekunden warten</German>
```
AirwavesMan: For the german string:

Du hast einen falschen Code eingegeben, warte %1 Sekunden.

icomrade:
Instead of making a new variable, why not just define dayz_UnlockTime  in configVariable.sqf? You wouldn't even need to modify the sqf file for door management at that point

oiad:
You need to modify the door management sqf to actually lock out the keylock till the timer is satisfied.

the variable harderpenalty was in the original door management, I dont know why it wasnt ported to here as well.

The only thing the variable does it make it exponentially increase if true or sets the timer to 5 seconds if false

SQF/dayz_code/stringtable.xml
oiad: Both added, thanks again @AirwavesMan

Issue #1854, opened 2017-01-05
Moving loot from veins and ammo crate to loot tables


AirwavesMan:
Can we move the loot from the ore veins and the ammo crates to the day_code loot tables? There are already loot tables with that name but we dont use them(e.g. ammo crate, dynamic debries. The loot for veins and the ammo crates are defined in the configs from these objects. I tried to move them to the missionfile and it worked only sometimes properly. 
It should work like the debries loot table which are in the remove.sqf I think. I know it is not a priority but a good benifit at all because many admins want to change these loot tables but they are no able to do so.

schwanzkopfhegel:
Maybe something else related to the ore veins:
For me and my players it seems like the veins are spawning nearly at the same spots. Are we able to randomize it a bit more?
It is possible to gain much "Gold" in short time just by searching some hot spots of the veins..

DeVloek:
@schwanzkopfhegel It's similar on Takistan, 99% of the veins spawn only in the north-east quadrant of the map, and most of them always in the same spots, +- 200m. On certain custom maps it is even worse, for example on Chernobyl the veins spawn in a long row along the edge of the southern forest.

oiad:
@AirwavesMan I have a working example for the ammo crates to loot tables, I just need to tidy it up, it MAY not be the best way to do it but it works great, I have it reading from the sniperrifiles loot group at the moment.

AirwavesMan:
@oiad That sounds good, perhaps we can use it for the ore veins as well?

oiad:
Okay so I have both of these moved to CfgLoot with no issues, it works great but I don't completely understand the loot chance etc so before making a pull request I'd like you to have a look over the loot tables side of it (I've never messed with loot tables) @AirwavesMan and @ebayShopper over the code if you have time.

https://github.com/oiad/DayZ-Epoch/commit/581c5ba4c14a992ba862252e019384efbceae885

ebayShopper:
The mine vein spawning position code hasn't changed since 1051 (HeliCrashArea is the same as (getMarkerSize "center") select 1):
https://github.com/EpochModTeam/DayZ-Epoch/blob/master/SQF/dayz_server/compile/spawn_mineveins.sqf#L3-L7 
https://github.com/EpochModTeam/DayZ-Epoch/blob/Release_1.0.5.1/SQF/dayz_server/init/server_functions.sqf#L463-L469

[...]

PR #1863, opened 2017-01-08
Move dze_supplyCrate and vein wrecks to CfgLoot

oiad:
Moves both the supply crates and vein wrecks to the CfgLoot tables so admins can customize.
Also modifies the main add item statement to allow support for weapons, magazines and backpack refunding/spawning instead of hard coded magazines.

Should be backwards compatible with the old style (I did test).

https://github.com/EpochModTeam/DayZ-Epoch/issues/1854


AirwavesMan:
Very good, it works like a charm

PR #1882, opened 2017-01-18
Update stringtable.xml

AirwavesMan:


ebayShopper:
Please also move this string down into the `dayz_epoch` package. it makes it easier to compare and sync the stringtable with vanilla updates when we keep Epoch specific strings in there.

AirwavesMan:
Done

PR #1884, opened 2017-01-21
fix epoch_tempKeys.sqf making massive arrays.

oiad:
epoch_tempKeys was making a massive array every time it was called due
to them setting the array size to the _ownerKeyId instead of count
_temp_keys_names.

For example: Call this function 10k times in testkit and see it lag the
client due to making a massive array over and over again.

```
"[[["9379"],[<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<null>,<n
"Test script finished. Code took 0.440002 seconds to run"
```

With fixed code:
```
"[[["9379"],["Yellow Key (65d0)"]],false]"
"Test script finished. Code took 0.000991821 seconds to run"
```

### DIFF
--- a/SQF/dayz_code/compile/fn_selfActions.sqf
+++ b/SQF/dayz_code/compile/fn_selfActions.sqf
@@ -5,7 +5,7 @@ scriptName "Functions\misc\fn_selfActions.sqf";
 	- [] call fnc_usec_selfActions;
 ************************************************************/
 if (dayz_actionInProgress) exitWith {};
-private ["_canPickLight","_text","_unlock","_lock","_totalKeys","_temp_keys","_temp_key_name","_temp_keys_names",
+private ["_canPickLight","_text","_unlock","_lock","_totalKeys","_temp_keys","_temp_keys_names",
 "_hasKey","_oldOwner","_hasAttached","_isZombie","_isHarvested","_isMan","_isFuel","_hasRawMeat","_hastinitem","_player_deleteBuild",
 "_player_lockUnlock_crtl","_displayName","_hasIgnitors","_menu","_menu1","_allowTow","_liftHeli","_found","_posL","_posC","_height","_attached",
 "_combi","_findNearestGen","_humanity_logic","_low_high","_cancel","_buy","_buyV","_humanity","_traderMenu","_warn","_typeOfCursorTarget",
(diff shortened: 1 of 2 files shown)


oiad:
@ebayShopper this has to be quite a lot better for performance since we call this function everytime fn_selfActions is refreshed.

SQF/dayz_code/compile/fn_selfActions.sqf
```diff
@@ -120,10 +120,15 @@ if (_inVehicle) then {
 			_temp_keys_names = _totalKeys select 1;	
 			_hasKey = _vehicleOwnerID in _temp_keys;
 			_oldOwner = (_vehicleOwnerID == _uid);
+			
+			{
+				if (_vehicleOwnerID == _x) exitWith {_temp_key_name = _temp_keys_names select _forEachIndex};
+			} forEach _temp_keys;
```
ebayShopper: Find is simpler to get an index. You can just replace this code below:
```sqf
(_temp_keys_names select (parseNumber _vehicleOwnerID))
```
with this:
```sqf
(_temp_keys_names select (_temp_keys find _vehicleOwnerID))
```
Same for the _characterID one. Then the forEach loops and _temp_key_name variable aren't needed.

oiad: @ebayShopper changes made

PR #1897, opened 2017-02-24
Extend epoch_tempKeys to be able to read from backpack as well.

oiad:


icomrade:
did you want to modify FN_selfActions so backpack keys can unlock vehicles as well?

ebayShopper:
You should make this the default behavior. It is only called from self actions and the function can be easily overwritten if toolbelt only is desired. That way we aren't needlessly checking another condition on loop.

```sqf
_items = items player;
dayz_myBackpack = unitBackpack player;

if (!isNull dayz_myBackpack) then {
	_items = _items + ((getWeaponCargo dayz_myBackpack) select 0);
};

{
	//check keys
} count _items;
```

oiad:
@icomrade I think it may be helpful to unlock from backpack as well, maybe a toggle for that

@ebayShopper I wanted a way to be able to differentiate between them (instead of them all jumbled together), a lot of scripts use epoch_tempKeys (external) so being able to use this with backpack support was my thinking.

icomrade:
I would just make it so vehicles unlock with backpack keys by default. Actually I don't even think a toggle is needed

PR #1912, opened 2017-03-19
german string update

AirwavesMan:


DAmNRelentless:
Please explain that change to me.
Why not "Ihr Name war %1, anscheinend starb sie an %2."?
And also "%1 Selbstmord"? Isn't that a bit weird? Why not "%1 beging Suizid" or "%1 beging Selbstmord"?
Another example: "%1 starb %2"

So please bring me up to date for the reasons you changed it.

AirwavesMan:
I made the chages because of how the death messages are build. I cannot use the first sentence because it will have the same words twice. For example: 

"Ihr Name war %1, anscheinend starb sie an an einer Schusswunde.."

The words and the punctuation mark are set twice. 

I agree with you  for "%1 Selbstmord"? I think it was a mistake from me ealier like the other change. I will update the other death causes as well. Take a look at it perhaps I missed something.

The last one is the same like the first one. Look at this line:

https://github.com/EpochModTeam/DayZ-Epoch/blob/d8828d9df1d14239b7148f28ce84a3dce19d5d89/SQF/dayz_code/compile/dze_deathMessage.sqf#L8

Basti890:
as the File is touched once there could be also made some more adjustments like:
`		<Key ID="str_actions_stats_hm">
			<English>Humanity</English>
			<Russian>Гуманность</Russian>
			<Spanish>Humanidad</Spanish>
			<Czech>Lidskost</Czech>
			<French>Humanité</French>
			<German>Menschlichkeit</German>
		</Key>`
The german word should be "Karma" or something like this. "Menschlichkeit" in this context is not common to use in german.

AirwavesMan:
@Basti890 I agree with you but "Karma" is not a appropriate word. Better is "Humanität" or "Barmherzigkeit" but these words are the same like Menschlichkeit. Everything sounds wrong . Ive asked many players for a better version of Menschlichkeit and all players think it should stay as it is.

DAmNRelentless:
Yeah, it's more "Humanität". Basti890 is partly right as it is kindly a karma system but as it is not karma in english it should either stay like "Menschlichkeit" or you choose "Humanität". I prefer the second one but that's the choice of anyone who edits the stringtable so not neccessary to edit.

In my opinion you should change a lot more of the german strings because the grammar is very horrible sometimes but I think first of all it's important to complete everything before improving it.

DAmNRelentless:
Line 7314: Change the "Momenten" to "Momentan"
Line 7326: Replace it with "Warte bis die vorherige Aktion abgeschlossen ist, bevor du eine neue beginnst!" as the current translation sounds really weird.

These are the changes I would suggest in the current section.

AirwavesMan:
@DAmNRelentless Honestly I dont have some kind of overview what strings are bad. I know a few are not that good. We could work together on a proper solution and then push it to master here. I have the current version of epoch in my repository. So feel free to push some changes to my repository and we can discuss it there and after that we push it to master.

PR #1916, opened 2017-03-22
Update player_addToolbelt.sqf

F507DMT:
player removeWeapon _item; //Need use BIS_fnc_invRemove, or player can dupe item

### DIFF
--- a/SQF/dayz_code/actions/player_addToolbelt.sqf
+++ b/SQF/dayz_code/actions/player_addToolbelt.sqf
@@ -58,7 +58,7 @@ if ((_item in ["ItemHatchet","ItemCrowbar","ItemMachete","ItemFishingPole","Item
 
 _isOk = [player,_config2] call BIS_fnc_invAdd;
 if (_isOk) then {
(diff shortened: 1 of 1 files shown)


icomrade:
you can only carry 1 of each tool-belt item, is there actually something you can dupe?

ebayShopper:
You didn't change anything. You just added a comment. Please explain how the dupe works so we can confirm if it exists and fix it if necessary. Thanks.

ebayShopper:
I see now what you were referring to. This script was not properly checking if the player still had the original tool before adding the new one. It would allow you to not have the original if you had any melee weapon on your back on this line:
https://github.com/F507DMT/DayZ-Epoch/blob/3a87e7e6d53c285267fa66bc36f9596140ddcd7f/SQF/dayz_code/actions/player_addToolbelt.sqf#L21

I fixed that error and simplified the script. It now uses invRemove at the end to double check that you still have the weapon.

Player_dropWeapon had the same problem. Fixed with 87abb20.

PR #1918, opened 2017-03-23
Add DZE_limitPlots so admins can limit plot poles to 1 per UID.

oiad:
This adds DZE_limitPlots where admins can restrict a player to only
having 1 plot pole. Disabled by default.

### DIFF
--- a/CHANGE LOG 1.0.6.2.txt
+++ b/CHANGE LOG 1.0.6.2.txt
@@ -1,4 +1,5 @@
 [NEW] Added a few loot positions for Namalsk and Takistan. @skigoggles
+[NEW] Added DZE_limitPlots so admins can limit plot poles to 1 per UID, disabled by default. @oiad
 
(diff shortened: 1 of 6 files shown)


AirwavesMan:
Thats a good addon. I have that on my server as well.

Cherdenko:
<German>Du darfst nur ein Grundstück haben, auf dem du der Besitzer bist.</German>

Cherdenko:
German String: Du darfst nur ein Grundstück haben, auf dem du der Besitzer bist.

DAmNRelentless:
I suggest "Du darfst maximal ein Grundstück besitzen."

oiad:
What's the difference?

DAmNRelentless:
The grammar is better.
It's the short form of "The maximum number of properties you are allowed to own, is one."

Actually I am reworking all german strings as you know already and that would more fit into the structure I am currently using.

PR #1923, opened 2017-03-26
Update fill_nearestVehicle.sqf

F507DMT:
() - condition
{} - code execution, heavier code

### DIFF
--- a/SQF/dayz_code/actions/fill_nearestVehicle.sqf
+++ b/SQF/dayz_code/actions/fill_nearestVehicle.sqf
@@ -15,7 +15,7 @@ if (!isNull _fuelTruck) then {
 
 _findNearestVehicle = [];
 {
-	if ((alive _x) && {_x != _fuelTruck} && {!(_x isKindOf "Man")}) exitWith {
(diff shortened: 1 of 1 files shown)


icomrade:
Lazy eval {} should be much quicker in this case since alive is faster than isKindOf and may be similar to a string comparison. Allvehicles may contain dead/destroyed objects, so I see no need to remove the lazy eval

ebayShopper:
Lazy evaluation is slower for `AND` conditions when they are usually true. That's because evaluating code is a little slower than evaluating a boolean. So if all or most of the conditions are usually evaluated anyway then lazy evaluation is slower. See:
https://community.bistudio.com/wiki/Code_Optimisation#Conditions

In this case this script is only executed when the player is looking at a fuel truck with the engine on that is a max of four meters away or a fuel pump with a running generator nearby. NearestObjects is sorted by distance, so the first returned object is usually going to be the player. In that case all three conditions will evaluate no matter what. The second or third returned object is usually the target vehicle, at which point this exits and does not check any other vehicles.

So in this case lazy evaluation should be slower, but it is a minor difference, since it is usually only checked two or three times max.

PR #1928, opened 2017-04-01
Update player_harvestPlant.sqf

F507DMT:
Added the ability to collect "FiberPlant": object="FiberPlant" createVehicle(position player);
fix many players collect 1 plant, and have many items.

### DIFF
--- a/SQF/dayz_code/actions/player_harvestPlant.sqf
+++ b/SQF/dayz_code/actions/player_harvestPlant.sqf
@@ -3,13 +3,14 @@
 	Usage: spawn player_harvestPlant;
 	Made for DayZ Epoch please ask permission to use/edit/distrubute email vbawol@veteranbastards.com.
 */
-private ["_isOk","_i","_objName","_started","_finished","_animState","_isMedic","_proceed","_itemOut","_countOut","_tree","_trees","_findNearestTree","_index","_invResult","_treesOutput","_text","_obj"];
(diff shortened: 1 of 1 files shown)


F507DMT:
I try not to use extra characters, spaces, to optimize the code.
Do I need to add the necessary localization myself?

oiad:
Prefer tidyness, it won't optimize it that much. Just makes it harder to read and debug if need be.

Yes, you need to add the localization, preferably in the epoch part of the stringtable

F507DMT:
ok! i'll know.
Done

icomrade:
when the code is compiled formatting and comments are removed, so readability and consistent formatting are preferred.

ebayShopper:
This won't work when the plant is not local to the player. That happens if you use createVehicle on the server, another client or relog so the locality changes. `Fn_getModelName` returns the model name with " remote" on the end like: `"p_fiberplant_ep1.p3d remote"`. It was the same issue with trees in 100f9f5.

`Fn_getModelName` needs to be modified to strip off the " remote" on the end of the return string if it is there. All three plant types should be added to the array so admins can spawn extra ones with createVehicle too if they want. If you know how to do it then go ahead, or I'll do it when I have time in a few days.

F507DMT:
Yes, for sure, did not pay attention.
Done!

F507DMT:
Yes, for sure, did not pay attention.

Added the ability to collect "MAP_pumpkin" and "MAP_p_Helianthus".
I left the searches for the model so that there were no problems with the planted ones "MAP_pumpkin" and "MAP_p_Helianthus".
Done!

PR #1934, opened 2017-04-07
Convert DZE_SafeZoneNoBuildItems to handle nested arrays for custom d…

oiad:
…istances per items.

This makes the DZE_SafeZoneNoBuildItems be able to handle nested arrays,
This allows you to set custom distances per item instead of having it
all the default distance.

E.g DZE_SafeZoneNoBuildItems =
["VaultStorageLocked","LockboxStorageLocked",["Plastic_Pole_EP1_DZ",1300]];

### DIFF
--- a/SQF/dayz_code/compile/dze_buildChecks.sqf
+++ b/SQF/dayz_code/compile/dze_buildChecks.sqf
@@ -106,7 +106,7 @@ _text = getText (configFile >> 'CfgMagazines' >> _item >> 'displayName');
 
 _buildCheck = call _checkClass;
 
-if (((count DZE_SafeZoneNoBuildItems) > 0) && (_buildCheck select 0)) then {
+if (_buildCheck select 0) then {
 	{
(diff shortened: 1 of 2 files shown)


SQF/dayz_code/compile/dze_buildChecks.sqf
```diff
@@ -12,6 +12,24 @@ if (!canbuild) exitWith {dayz_actionInProgress = false; format[localize "STR_EPO
...
+	private ["_checkOK","_distance"];
+
+	_checkOK = false;
+	_distance = DZE_SafeZoneNoBuildDistance;
+
+	{
+		if (typeName _x == "ARRAY") then {
+			if (_x select 0 == _classname) exitWith {_checkOK = true; _distance = _x select 1;};
+		} else {
+			if (_x == _className) exitWith {_checkOK = true};
```
ebayShopper: Using exitWith here only exits the Then { } else { }; scope, not the count loop, so this loop is always just checking the last element in the array right now.

PR #1955, opened 2017-06-21
Fix Corepatch AS50 magazine spawn

AirwavesMan:
The Corepatch broke some of the AS50 magazine spawns. Im not sure why it happend sometimes and sometimes not.

This message showed up sometimes: No entry 'bin\config.bin/CfgMagazines-5Rnd_127x99_as50_CP'.

The CorePatch use that: 

https://github.com/Goliath86/CorePatch/blob/e0d06f7086fb0e3b3e6b7dc22beeca3b087967be/CorePatch_CIT_14888/config.cpp

This commit should fix that error message and the AS50 name as well.

### DIFF
--- a/CHANGE LOG 1.0.6.2.txt
+++ b/CHANGE LOG 1.0.6.2.txt
@@ -48,6 +48,7 @@
 [FIXED] Readded crossbow reload sound
 [FIXED] Using setDamage or setHit to kill a player via script will no longer cause a double death.
 [FIXED] The dramatic recoil camera shake effect from a nearby bullet hit is now reset correctly instead of remaining permanent.
+[FIXED] Added temporary fix for missing AS50 ammo error with beta branch core patch. #1955 @AirwavesMan
 
 [NOTE] Fixes below are included in the mission file and server pbo as part of server package 1.0.6.1A (March 10th 2017)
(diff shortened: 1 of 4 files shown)


ebayShopper:
This happens when a client on the server is using the beta branch of A2OA. If you have an AS50 in your loot tables or traders then that client will spawn it with the new magazine type, because it is first in the magazines array in the beta branch. The other stable branch clients will get the magazine type does not exist error.

Your change makes it worse for stable clients because it appears each magazine type is checked when the weapon is spawned. I got the error with your change even though I only spawned this weapon with the normal ammo.

Instead you will need to define `5Rnd_127x99_as50_CP` in CfgMagazines. Please also add a comment with it that says:
```
//Remove after A2OA stable branch exceeds Version 1.63.131129
```

Also please add to the changelog:
```
[FIXED] Added temporary fix for missing AS50 ammo error with beta branch core patch. #1955 @AirWavesMan
```

This is fine for a temporary solution.

You may also want to add a kick for clients running the beta branch before they load in your server, since they may have other compatibility problems too. We should not add a kick to the mod due to an inevitable lapse between an A2OA stable branch update and mod patch.

AirwavesMan:
All changes are done.

PR #1984, opened 2017-10-01
Fix completely broken dze_buildChecks

oiad:
Revert https://github.com/EpochModTeam/DayZ-Epoch/commit/d750f576eda3949f365b2fc97316b4faec7327d7 since it's using the wrong distance check (_buildCheck select 1 is needed for the nested array check)
Revert https://github.com/EpochModTeam/DayZ-Epoch/commit/bbb7da7c3c6572e1591189600ef55bf8b7eb2d71#diff-0cdf3d56b40b62e92b498e7362874f34 I broke this so that it would never properly exit, it's now done properly


icomrade:
did you reference https://github.com/EpochModTeam/DayZ-Epoch/commit/d750f576eda3949f365b2fc97316b4faec7327d7 erroneously?

PR #1985, opened 2017-10-01
Structured keyboard.sqf

DAmNRelentless:
I recently messed around alot with this file. I just structured it a bit to make it easier to read and to find things.

### DIFF
--- a/SQF/dayz_code/compile/keyboard.sqf
+++ b/SQF/dayz_code/compile/keyboard.sqf
@@ -2,331 +2,337 @@
 
 #include "\ca\editor\Data\Scripts\dikCodes.h"
 
-_dikCode = _this select 1;
+_dikCode    = _this select 1;
 _shiftState = _this select 2;
-_ctrlState = _this select 3;
-_altState = _this select 4;
-_handled = false;
+_ctrlState  = _this select 3;
+_altState   = _this select 4;
+_handled    = false;
 
 if (isNil "keyboard_keys") then {
-    _cancelBuild = {
-		DZE_cancelBuilding = true;
-		call dayz_EjectPlayer;
-		_handled = false;
-		if (r_player_dead) then {_handled = true;}; // Disable ESC after death
-    };
-	_dze_f = {
-		 if (!_ctrlState && !_altState) then {DZE_F = true;};
-	};
-	_dze_q = {
-		 if (!_ctrlState && !_altState) then {DZE_Q = true;};
-		 if (!_ctrlState && _altState) then {DZE_Q_alt = true;};
-		 if (_ctrlState && !_altState) then {DZE_Q_ctrl = true;};
-	};
-	_dze_z = {
-		 if (!_ctrlState && !_altState) then {DZE_Z = true;};
-		 if (!_ctrlState && _altState) then {DZE_Z_alt = true;};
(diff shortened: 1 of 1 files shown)


SQF/dayz_code/compile/keyboard.sqf
```diff
@@ -2,331 +2,337 @@
...
+  };
+  _pistol = {
+    3 call dz_fn_switchWeapon;
+    _handled = true;
+  };
+  _melee = {
+    //Second rifle if DZE_TwoPrimaries = 2 (dayz_onBack)
+    4 call dz_fn_switchWeapon;
+    _handled = true;
+  };
```
DAmNRelentless: DIK_4 which handles the throwable is not used by default so I commented the throwable code

SQF/dayz_code/compile/keyboard.sqf
```diff
@@ -2,331 +2,337 @@
...
-        };
+  };
+  _forcesave = {
+    if (diag_tickTime - dayz_lastSave > 10) then {
+      call player_forceSave;
     };
-	_surrender = {
-		call player_surrender;
-		_handled = true;
+  };
```
DAmNRelentless: Same for the forcesave2, it's not used by default

SQF/dayz_code/compile/keyboard.sqf
```diff
@@ -2,331 +2,337 @@
...
+    if (_ctrlState && !_altState) then {DZE_Z_ctrl = true;};
+  };
+  _autoRun = {
+    if (!dayz_autoRun) then {
+      dayz_autoRun = true;
+      dayz_autoRunThread = [] spawn {
+        _weapon = currentWeapon player;
+        while {dayz_autoRun} do {
+          //SurfaceIsWater doesn't work for ponds
+          //Check Weapon to detect ArmA action (dayz action is handled in dz_fn_switchWeapon)
```
DAmNRelentless: This will probably take a lot of performance due to several statements on an if in a loop. May increase the uiSleep?

SQF/dayz_code/compile/keyboard.sqf
```diff
@@ -2,331 +2,337 @@
...
+  [actionKeys "Chat", _noise] call _addArray;
+  [actionKeys "User20", _journal] call _addArray;
+  [actionKeys "Diary", _journal] call _addArray;
+  [actionKeys "NetworkStats", _journal] call _addArray;
+  [actionKeys "LeanLeft", _build_left ] call _addArray;
+  [actionKeys "LeanRight", _build_right ] call _addArray;
+  [actionKeys "PersonView", _build_camOnOff ] call _addArray; //Camera Mode
+  [actionKeys "GetOver", _build_str8OnOff ] call _addArray; //V
+  [actionKeys "ForceCommandingMode", {DZE_5 = true;_handled = true;}] call _addArray;
+
```
DAmNRelentless: Added comment so users can easier see that these are the keys that are blocked

icomrade:
I don't really see this as necessary, I'd prefer to leave it untouched for easier comparison with the vanilla code in case of another major update. Also, switching from tabs to spaces for indentation goes against the standard formatting practices this project uses.

DAmNRelentless:
I used tabs. Probably the formatting of the editor I am using is different (using Atom).

PR #1998, opened 2017-12-02
Update player_death.sqf

BigEgg17:
Replaced titleText with new default dayz_rollingMessages.


ebayShopper:
TitleText was left here intentionally so it doesn't overwrite death messages when **DZE_DeathMsgRolling = true;**

PR #2008, opened 2017-12-23
Update Russian translations

dreamforceinc:


dreamforceinc:
"STR_EPOCH_KILOTOBACCO_DESC"
"STR_EPOCH_KILOBLACKTEA_DESC"
"STR_EPOCH_TOBACCO_LEAF_DESC"
Is this not a description?

dreamforceinc:
And one more question: "Oil barrel" - what is meant? Engine's oil or a petroleum?

ebayShopper:
Yes, they are descriptions, but since they are incomplete sentences they should not have periods.

I know the other descriptions in the stringtable are mixed right now. Some are fragments and others are complete sentences. Some have periods and some do not, but no period for fragments is a good convention to follow.

The "Oil barrel" is for petroleum.

dreamforceinc:
Can I get examples of that sentences? Maybe I will need to change the translation.

ebayShopper:
I don't know which type of oil @AirwavesMan wanted this to be.

It's the same model and picture as `ItemFuelBarrel`
https://epochmod.gamepedia.com/A2Epoch:Fuel_Barrel

"Fuel" is a less specific word. In this case we are just using it to describe a source of energy which powers vehicles.

I assume "Oil" in this case refers to crude oil or oil used in ships/furnaces. It is not important, just make sure it is different from this translation so they can be distinguished in the gear menu:
https://github.com/EpochModTeam/DayZ-Epoch/blob/master/SQF/dayz_code/stringtable.xml#L14561-L14563

dreamforceinc:
ok, then I'll leave it as it is.

PR #2026, opened 2019-01-04
Update player_countMagazines.sqf

F507DMT:
Variables("_item","_val","_max") get lost in other scripts, causing errors.

### DIFF
--- a/SQF/dayz_code/compile/player_countMagazines.sqf
+++ b/SQF/dayz_code/compile/player_countMagazines.sqf
@@ -4,7 +4,7 @@
 	return all player magazines with ammo count
 */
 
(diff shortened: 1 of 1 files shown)


Treuce:
Would it not be easier to just rework that script to skip the gear dialog creation aswell?

F507DMT:
i used the _item variable in the script, then saving.
the script give empty _item, in the middle of work.

I think _item, a frequent name for a variable. In addition, it has been added to player_countMagazinesWBackpack.sqf.

Treuce:
It would need to be in the private ["_dialog","_created","_magazineArray"]; and that would fix that issue I believe, but I find this script annoying because it opens the gear dialog which could be done in an other way without having to create a dialog.

PR #2032, opened 2019-02-08
Update player_switchWeapon.sqf

F507DMT:
It is forbidden to change weapons, when player swimming.


oiad:
Why forbid them changing weapons while swimming? Is it to prevent an exploit or similar?

Also, in future please format your code neatly with spaces/tabs
i.e:
`if(dayz_isSwimming)exitWith{};` should become:
`if (dayz_isSwimming) exitWith {};`

It just makes it easier to follow/read

F507DMT:
Sorry, I'm just writing code without spaces.
The character floats, then rises into the water, changes weapons, and floats again

PR #2066, opened 2020-02-20
Fix parachute bug

ch0c:
Players can spam certain keys to make their parachute bypass its max speed.
Attached file will pull the player towards the ground when they reach a max speed of 20km, the faster they go, the harder it will pull them. 

[fn_halo.txt](https://github.com/EpochModTeam/DayZ-Epoch/files/4229524/fn_halo.txt)


AirwavesMan:
Added with https://github.com/EpochModTeam/DayZ-Epoch/commit/2a41d122f66292f87f8280215c9d72de49ca7706

PR #2070, opened 2020-03-18
Check permaLoot when building_spawnLoot deletes Reammobox

Schalldampfer:
I fixed an issue that permaLoots get deleted by Loot Spawner


AirwavesMan:
This will not fix the problem. The perma loot variable is set server side only. We probably could make the variable public but this would affect the network traffic more. I rather would stick to how it is at the moment.

Schalldampfer:
It's fixed in this way in my server, and it's only a check of a variable.
Of course it may affect traffic, but I don't think there're many permaLoots which must have the public variable. 
I thinks only a few reward boxs in static missions need it  when they're placed on exsiting buildings.

ArmaWizard:
Hi,

Does Epoch spawn reammo boxes by default that are getting deleted, is that the issue? Or is it only on your static missions @Schalldampfer 

Greetings,
Merlijn

Schalldampfer:
I don't know issues about DayZ Epoch's loot boxes, I only know boxes in mission systems have permaloot.

ArmaWizard:
If it's not a bug regarding Epoch itself I don't quite see the need to make an exception so a third party scripts functions correctly.

PR #2100, opened 2021-01-23
Integrate clickActions

Schalldampfer:
The clickActions part of https://github.com/oiad/deployAnything
I've seen this script before the repository, but I don't know the history of it.


AirwavesMan:
The rightclick action is a community script that is very esay to install. There is no need to be added.

Schalldampfer:
I got it.

PR #2120, opened 2021-03-29
Update stringtable.xml

dreamforceinc:
Revert back the russian translation of 'humanity'.


AirwavesMan:
@dreamforceinc Do you plan more updates in the next days? If yes push them until the 11th april otherwise I cannot add them to release.

dreamforceinc:
>Do you plan more updates in the next days? If yes push them until the 11th april otherwise I cannot add them to release.

I think that's all.

PR #2132, opened 2021-05-03
Update player_emptyContainer.sqf

F507DMT:
1. dup's attempt
2. dayz_actionInProgress=false; - missing afther exitWith


AirwavesMan:
I will apply a different fix

PR #2151, opened 2021-08-02
Update fn_selfActions.sqf

looter809:
Fix players not being able to lock a safe that they do not have the combination for.


worldwidesorrow (commented):
We still need to check that the cursor target is a vault, locked or unlocked. Otherwise the re-enter combination self action will be added for other objects. So we can use this:

if (_isClose && !keypadCancel && {(_typeOfCursorTarget in (DZE_LockedStorage + DZE_UnLockedStorage))  && {_characterID != "0"}}) then {

looter809:
Oh that’s very true. Good catch

looter809 (commented):
Updated as worldwidesorrow suggested

PR #2158, opened 2021-11-07
Add missing urals (credit to Kitty on Discord)

ZzBombardierzZ:


ZzBombardierzZ:
I just noticed that maybe I should have done all INS urals above the CDF urals, as that is what seems to be the pattern

PR #2162, opened 2021-12-05
Update server_verifySender.sqf

ZzBombardierzZ:
Fixes issue with objects being placed over water not being saved to hive. 
_objPos's position is gotten through getPosATL in the modular_build.sqf and player_build.sqf files, and by not comparing to the player's ATL position causes server_verifySender to throw the error "%1 error: Verification failed, player is too far from object. Distance: %2m/%3m limit PV ARRAY: %4"

### DIFF
--- a/SQF/dayz_server/compile/server_verifySender.sqf
+++ b/SQF/dayz_server/compile/server_verifySender.sqf
@@ -20,7 +20,7 @@ _exitReason = call {
 	//If object or player is null distance returns 9999+
 	//If object or player was moved with setPos on client, position takes a second to update on server
 	//Coordinates can be used in place of object
-	if (_objPos distance _player > (Z_VehicleDistance + 10)) exitwith {
+	if (_objPos distance getPosATL _player > (Z_VehicleDistance + 10)) exitwith {
 		format["%1 error: Verification failed, player is too far from object. Distance: %2m/%3m limit PV ARRAY: %4",_function,round (_objPos distance _player),Z_VehicleDistance + 10,_params]
 	}; 
 	if (_index < 0) exitwith {


AirwavesMan:
The whole building system got a big overhaul in 1.0.7.1. and change will no longer be needed. But it is possible that we need it for the other checks which server_verifySender is used for. I will take at look at it.

ZzBombardierzZ:
Ok, I may be wrong but I just looked into the new build system and I still think that building over water would be an issue without the change I made, but I could be wrong. 

The issue is not the object position, as the _tmpBuilt position as it is and in 1.0.7.1 is converted to ATL before sending to the server. The issue is when requesting to build something to the server sends the player as an object, and when the server verifies the object position with the distance of the player, it is wrong unless you use getPosATL on the player object.

Here's a screenshot of what I mean: https://ibb.co/d40nyk3   and remember that the object's position is sent as ATL which is correct. In the screenshot you can see that I am 53m over the bottom of the ocean, because when using distance from the ATL position to the actual position it returns 53. Meanwhile, server_verifySender gets upset if the distance between the obj and the player is greater than the vehicle trader distance (Z_VehicleDistance) + 10.

AirwavesMan:
I understood what you mean. The actual changes of the modular_build are not live so far. Im also not able to reproduce the bulding problem with building on water but the new modular_build will handle that anyway. I just want to make sure that the position check will not be a problem with the other functions. It shoud not be a problem but better to check that twice since it is a critical function.

ZzBombardierzZ:
To reproduce in 1.0.7 (and possibly 1.0.7.1), go out to sea at least 50-100m from shore -I used about 20 2x2 metal floors- and try to build a base. The _tmpBuilt stays for the restart but if you check the server.rpt while building, the verifySender throws publish failure logs, and the objects are not saved to the database, thus not there after restart.

I was also concerned that it may cause issues with other functions but I don't really think it should. Of course, you would know better though.

AirwavesMan:
Fixed with https://github.com/EpochModTeam/DayZ-Epoch/commit/5dafe02553a6c9a805c4c8d3aced472563959a8c

PR #2177, opened 2022-04-27
Fix CfgVehicles\Buildables\Sandbags.hpp

dreamforceinc:


dreamforceinc:
And I need edited config.bin for more testing.

PR #2191, opened 2023-07-25
Fix prefix missing property and pboprefix showing as .txt

ZzBombardierzZ:
My last pull request was somehow missing the PREFIX property in the dayz_server.pbo and PBOPREFIX property was showing as a txt file. Should be fixed this time. My apologies for not double-checking the last one.


ndavalos:
Having $PBOPREFIX$.txt is normal, it depends on which program you are using to make your pbos, some require $PBOPREFIX$, some require $PBOPREFIX$.txt, it's a good idea to include both since you never can tell what people are going to use to make their pbos.

ZzBombardierzZ:
> Having `$PBOPREFIX$.txt` is normal, it depends on which program you are using to make your pbos, some require `$PBOPREFIX$`, some require `$PBOPREFIX$.txt`, it's a good idea to include both since you never can tell what people are going to use to make their pbos.

I encourage you to do a pull request with the correct format in dayz_server.pbo that you are mentioning. I fear I may mess it up again if I do it, my PBO manager & Mikero's PBO packer were both giving me issues the past couple of days.

Issue #2187, opened 2022-07-13
Epoch 1.0.7.1 Bugs


AirwavesMan:
Currently reported bugs:

- Maintain the Virtual Garage with Coins does not work properly, this affects also the GUI. Players should just relog to get their coins back.
- A wooden base with wood walls and floors does not get properly recognized as shelter so it will rain through and the EVR-Storm can hit players inside.
- The Boggle is missing in the Safe Objects array so it will not stay after a restart.

We will prepare a hotfix file soon for the missionfiles.

ZzBombardierzZ:
Also it seems https://github.com/EpochModTeam/DayZ-Epoch/pull/2162 didn't fix building over water in my testing. It seems like the distance calculation when placing or removing an object over the water is still funky. 

`11:46:45 "Server_PublishObj error: Verification failed, player is too far from object. Distance: 46m/40m limit PV ARRAY: ["0",50c90f00# 1086282: sign_one_leg_h.p3d REMOTE,[193.144,[1032.23,447.776,55.2396],"76561198068668XXX",[[-0.2274,-0.973801,0],[0,0,1]]],[["76561198068668XXX",[66,111,109,98]]],B 1-1-B:1 (Bomb) REMOTE,"ıēŏĹĺĕŁłħŭŔńĈ"]"` <- when placing a plot right next to myself over the water.
Images: https://imgur.com/a/4gDmehv

Victor-the-Cleaner:
Hi @ZzBombardierzZ, this is definitely a weird one. When I click through to your image, it looks like you are standing in the sea (maybe from a teleport?) but not swimming. The modular_build script prevents building while swimming anyway.  One thing you could try is using the editor, place a few large metal floors at that location, allowing for about +1.8m at high tide, then add that to your custom map file.  Fly out there, then start building, and eventually remove the original floors from the map file?
Then you can go crazy and build the ultimate Bond villain seabase and take over the world.

All the testing for building on/over water was done the way a regular player would have to do it, so first building a pier out over the water and going from there.  Also bear in mind that certain objects that are flagged as not allowed below ground mean they are not allowed below the water, including the plot pole.  But it will tell you when the time comes.
Are you able to try this and let us know if that works? Thanks.

ZzBombardierzZ:
[...]

PR #2193, opened 2024-11-18
Update server_verifySender.sqf

ZzBombardierzZ:
Fixes building and removing objects over deep water - https://github.com/EpochModTeam/DayZ-Epoch/issues/2187#issuecomment-2483699804


AirwavesMan:
We will handle this with the upcoming update.

